PR #168, opened 2015-02-21
Rect test

WalliePL:
Test rect without alignment for negative height and widht. Tests failing in intersects methods - probably it's a bug in code.


AdamsLair:
Hi WalliePL and thanks for contributing :)

I have merged your Pull Request, but unfortunately had to make a large number of changes before committing the merged version back to GitHub. For the future, please keep the following in mind:

- The Duality repo uses tabs instead of spaces for indenting code.
- You tests should be medium to large-scale and each test for a number of related concepts. Don't implement tests that only check a single method.
- Your tests should be implemented as blackboxes wherever possible. Avoid using the TestCases attribute or similar when possible.
- Make sure to not accidentally commit temporary files, like in this case the testing reports xml file.
- In general, try to keep your tests a little more on the "macro" scale. They are really, really fine-grained right now. This may be good for other projects, but I'd try to avoid this for Duality. Test for functionality, not for API!

Think of a testing class like of a maintenance guy: Do you really want him to explain every single step he does, each time he does it? Probably not. What you really want is a guy that simply does his job and is able to simple tell you when something is broken, so you can investigate.

That said, there are some great things you definitely should keep up:

- Your overall coding style nicely matched the Duality style.
- You really thought of every possible way to break Rect and actually managed to do so in multiple cases, revealing two bugs that are now fixed.
- You took care of pulling from the newest Duality commit before sending in your Pull Request to keep merging effort on my side minimal. It also merged without problems, could easily have been an automatic one.
- You picked a small, manageable, non-critical task to solve. This is actually far more useful than jumping on the biggest issues right from the beginning!

Regards,
Adam

WalliePL:
Sorry for delay in answering. Thank You for giving me above informations! I will try to keep them in mind while developing new test or features :) Currently i have some time issues (a lot of work) so i'm not promising that new test will be soon. I have started work at Range class tests and i found some unused methods (lerp, normalize). Should i/you do something about it? If there are not used is there any reason to keep them?

PR #200, opened 2015-07-05
Fixed build on MonoDevelop

RockyTV:
This shouldn't break anything since the output assemblies were not renamed.


AdamsLair:
Since I didn't really dig into the topic yet and am not yet sure what this does or why it does it, I'll just leave this open as a note to myself for future consideration. Will revisit later.

PR #201, opened 2015-08-02
Added Windows 10 friendly name

RockyTV:
Just a tip: Duality should be targeted for Windows 10 too, otherwise running Duality on Windows 10 makes it think it is running on a Windows 8 environment.

See https://msdn.microsoft.com/en-us/library/windows/desktop/dn481241(v=vs.85).aspx


AdamsLair:
Plain and simple improvement - merged it!

As far as targeting goes, I'll probably wait with this until I have a Windows 10 machine for testing. As long as there aren't any issues from Duality thinking it's on Windows 8, this doesn't seem critical. In any case, thanks for the hint!

PR #215, opened 2015-09-30
Allow empty directories with the same name as the new project to be used as the project directory

### LINKED ISSUE #214: Empty directories with the same name as the project name should be valid
- Create a folder in `Documents\Duality Projects\`, name it TestProject.
- Open up Dualitor, and choose to create a new project.
- Name it TestProject, and select the recently created folder.
- Dualitor will throw an error saying that the new project target directory already exists.

Instead of just saying the directory isn't valid, Dualitor should check if the target directory isn't empty. If it is empty, the directory should be valid.

Should be pretty easy to implement.

RockyTV:
Should fix #214


DualityEditor/Forms/NewProjectDialog.cs
```diff
@@ -148,13 +148,16 @@ protected bool CheckInputValid()
...
 			if (Directory.Exists(targetDir) || File.Exists(targetDir))
 			{
-				this.textBoxFolder.BackColor = errorColor;
-				MessageBox.Show(
-					GeneralRes.Msg_CreateProjectErrorTargetExists_Desc,
-					GeneralRes.Msg_CreateProjectErrorTargetExists_Caption,
-					MessageBoxButtons.OK,
-					MessageBoxIcon.Warning);
-				return false;
+				if (Directory.EnumerateFiles(targetDir, "*", SearchOption.AllDirectories).Any())
```
ilexp: This will throw an Exception in case the `targetDir` path is actually a file path. You'll need to have a special case for that, so it will just display the error message without doing the directory-files check. Maybe split up the code, so you have one block for existing directories (with your files-check) and one for existing files (without)?

ilexp:
Looking good, except for the one thing I commented on. Waiting for your fix on that before merging.

RockyTV:
So, just split
```
if (Directory.Exists(targetPath) || File.Exists(targetPath))```

?

ilexp:
> So, just split
> ```if (Directory.Exists(targetPath) || File.Exists(targetPath))```

Yep. Split and add your emptiness-check only on the directory part.

RockyTV:
Alright, everything should work now.

PR #232, opened 2015-10-25
Flap or die sample

SirePi:
Full game, highscore management


ilexp:
Would like to merge this, but apparently, [the project fails to build](https://ci.appveyor.com/project/ilexp/duality/build/AppVeyor%20master%20Build%2097). Something appears to be wrong with the solution or project files.

> Error parsing the nested project section in solution file. A project with the GUID "{A16AF3AA-B690-4ADD-90DC-D5A90190AD0E}" is listed as being nested under project "{9D215950-8E34-4070-914F-7B7D8A54ED6C}", but does not exist in the solution.

Investigating the .sln file, [this line](https://github.com/SirePi/duality/blob/flap-or-die-sample/Duality.sln#L93) seems weird - why is there no EndProject?

SirePi:
Ah.. might have broken something with my last merge.. let me check

ilexp:
> Ah.. might have broken something with my last merge..
>
> EDIT: No, it all works and compiles correctly on my side

[A Project without an EndProject](https://github.com/SirePi/duality/blob/flap-or-die-sample/Duality.sln#L93) still looks broken to me. Maybe it just works in a certain compiler / environment setup?

Anyway, thanks for the fix! Will look into this later this day for the merge :)

SirePi:
Seems all working now. Still don't know how it was working on my end..

ilexp:
Merged and added some Tweaks. Thanks! :)

Some notes for the future:

- Don't add Duality settings (AppData, UserData, etc.) to the Content directory of samples, because they're not included in sample packages anyway.
- The content of this sample is awesome :smile: However, future samples really need to be smaller in file size overall, because git always has the full history of all files and can never get rid of it. 

---------

I probably won't merge future samples of this file size into the main repo, but would prefer them in distinct sample repositories like the one @hsnabn has set up. Remember: *Everyone* is forced to pull the **full history** of **all files** in this repo. Five MB doesn't sound like much, but it stacks up quickly.

The last paragraph is a bit of a note to myself, I should've thought about this earlier. ^^ Will add this to the main issue.

SirePi:
Ah.. my bad ^^ The original png is only 150Kb.. I didn't realize it would become that big once in resource form.

PR #254, opened 2015-12-01
Target Windows 8.1 and 10 on app.manifest

RockyTV:
This is required so Duality displays the correct Windows version instead of displaying the version as Windows 8.

Source: https://msdn.microsoft.com/en-us/library/windows/desktop/dn481241%28v=vs.85%29.aspx

### DIFF
--- a/DualityEditor/app.manifest
+++ b/DualityEditor/app.manifest
@@ -4,4 +4,12 @@
       <dpiAware>false</dpiAware>
     </asmv3:windowsSettings>
   </asmv3:application>
+  <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+    <application>
+      <!-- Windows 10 -->
+      <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>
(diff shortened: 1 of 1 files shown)


ilexp:
Small and self-contained PR - thanks! :)

Will need to evaluate what implications this change will have on the overall runtime environment, and whether internal changes are required to account for them. Will report back and keep this PR open until then.

ilexp:
Apparently, the `supportedOS` manifest file entries are used by Windows to determine compatibility settings that are applied. The Windows version that is returned by WinAPI or .Net is one of them, but not necessarily the only one, i.e. adding these settings might change runtime behavior on Windows 10.

I wasn't able to find a description of what exactly will be different depending on the presence or absence of the appropriate `supportedOS` entry. Duality seems to work fine on Windows 10 right now and I don't want to introduce any kind of incompatibility by changing environment settings - so this PR seems a bit risky to me at this point.

I'm closing this for now, but will revisit the issue at some point in the future. (Potentially when the dust on the Windows 10 release has settled and we all have a testing machine for that.)

Issue #231, opened 2015-10-24
Allow adding Components to multiple selected GameObjects


hsnabn:
Right now, you can only add components to one GameObject at a time. It gets tiring when, for example, you change a Component class's namespace and have to re-add it to different GameObjects.

It would be much easier when you can add Components to selected GameObjects in the same way as it is done right now.

ilexp:
Just to clarify: You're asking for an "Add a new Component of this Type to all selected GameObjects" editor feature, but **not** an engine core feature that would allow a Component to belong to multiple GameObjects at once - right?

I agree with the first option.

hsnabn:
Yep, the first option was what I was talking about.
I actually don't understand that second option entirely, actually.

ilexp:
> EDIT: A small thing I didn't want to create an issue for: the wiki could use an update. Duality 2.0 has been released, and there recently was someone with an issue with the wiki examples and Duality 2.0, so it would help newcomers if the wiki was up-to-date.

Fixed. Next time, please use a distinct issue though ^^

hsnabn:
I'll keep that in mind next time.

ilexp:
Done, see the referenced Pull Request above. Closing this.

PR #260, opened 2016-01-03
Added the ability to add components to multiple selected GameObjects

deanljohnson:
What I had to do:

1) Removed the check for a single selection when deciding if the "new" context menu item should be shown
2) Looped over the SelectedNodes property rather than just SelectedNode. This forced me to track the added nodes so that we could select them later. Selecting them inside the foreach loop would have modified the enumerable.

I checked and verified that this does not overwrite Component's if they already exist on a GameObject.

### DIFF
--- a/DualityEditorPlugins/SceneView/Modules/SceneView.cs
+++ b/DualityEditorPlugins/SceneView/Modules/SceneView.cs
@@ -747,7 +747,7 @@ namespace Duality.Editor.Plugins.SceneView
 			bool multiSelect = selNodeData.Count > 1;
 			bool gameObjSelect = selNodeData.Any(n => n is GameObjectNode);
 
-            this.nodeContextItemNew.Visible = gameObjSelect || noSelect;
+			this.nodeContextItemNew.Visible = gameObjSelect || noSelect;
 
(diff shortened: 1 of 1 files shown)


ilexp:
Didn't have the time for a thorough review yet, but will certainly look into your commits later! 

One thing I noticed though was your usage of spaces for indentation: Please change your code to use tabs instead, as specified in the (admittedly short and unpolished) [Contribution guide](https://github.com/AdamsLair/duality/wiki/How-to-Contribute):

> Always use Tabs for indentation, but Spaces for alignment.

ilexp:
Referencing issue #231 which is addressed by this.

deanljohnson:
Sorry about the spaces - my IDE was replacing tabs automatically it seems. Should be fixed now.

ilexp:
I've tested and merged your PR into the master branch. Unfortunately there were some issues in practice (fixed now), which you can read about [in my latest commit log](https://github.com/AdamsLair/duality/commit/3786c75dadb62ea2b336eed1034f9564652bf575).

(That aside, your code is in now and your PR was overall well executed. Properly done in its own branch, small and self contained, well documented, no unnecessary stuff, sticking to the existing style. Makes it a lot easier to handle contributions.)

Issue #261, opened 2016-01-04
Keep the dragging of Rigidbodies between objects from adding the default circle


deanljohnson:
Currently, if you drag a Rigidbody from one GameObject to another, it will add the default circle shape to the rigidbody. Steps to reproduce:

1. Create two empty GameObjects, A and B
2. Add a Rigidbody to object A
3. Replace the default circle collision shape with something else.
4. Drag and drop the Rigidbody and Transform from object A onto object B

Now if you look in the Rigidbody editor, you should see that it kept the shape we replaced the default shape with, but it re-added the default circle.

I'm assuming that this is because the Rigidbody is re-created rather than moved, and during it's construction the default circle is added. However, I also noticed that cloning a GameObject does not add back-in that default circle, so the functionality already exists somewhere to re-create the Rigidbody without adding that default circle.

ilexp:
Thanks for reporting! This is probably a minor bug in editor-related code.

deanljohnson:
I've tracked down the bug, but can't think of a good solution so I'm posting my findings here to see what other people think.

The problem occurs here:
https://github.com/AdamsLair/duality/blob/5f0f69127e09469c722e54f2d1dc2f6f816f9f2a/Duality/Components/Physics/RigidBody.cs#L1154

In the case of a drag-and-drop operation, OnInit is called with context set to AddToGameObject, and we are in the editor, so the RigidBody is assuming that it is a new RigidBody, which is not the case in a drag-and-drop operation.

At first I naively thought to check if the RigidBody's shape count was 0, but that'd cause the same bug if I was drag-and-dropping a RigidBody that I intended to have no shapes.

There could be a new InitContext, such as InitContext.Construction, but that clutters the InitContext enum even more, which just causes more problems for issue #136, so that's not a good fix. Perhaps the fix for this bug will stem from a fix to the InitContext issue.

We could add the default circle in the event handler for creating a new RigidBody. I haven't looked at that code yet so I'm not sure if that's going to provide a clean solution. Furthermore, I could see other Component's needing to know whether or not they are a completely new instance, so whatever solution is developed for this bug should address this issue for other Component's as well.

ilexp:
Very good analysis of the problem! I would see the solution to be implemented within the editor, rather than the core itself, since this is actually an editor feature which (unfortunately) was previously implemented in the core.

[...]

PR #262, opened 2016-01-07
Initialize Component Editor Actions

deanljohnson:
Implemented support for EditorActions that are responsible for initializing Component's specifically for use within the editor. 

These editor actions are called from within the CreateComponent UndoRedoAction. They were placed there because they would otherwise need to be called and managed by any third-party plugin that was making use of them, but I believe the usefulness of editor actions largely stems from their ease of integration between third-party and Duality code. And since CreateComponent is an editor related action anyways, it seems acceptable to have them placed there. If this becomes a point of dispute, it is trivial to move the calls somewhere else.

The change to the default value of Rigidbody.shapes is due to null reference exceptions that are thrown due to the change in initialization process.

This PR fixes issue #261.


deanljohnson:
Should the checks for initialization functions be removed from CreateComponent? Doing so would force plugins to check for and execute these functions on their own, which requires plugin developers do more work. Also, by having that code duplicated across multiple plugins it's more likely that bugs can creep in. 

With the current implementation, using this functionality just requires the right EditorAction to be defined, and that the plugin uses the CreateComponent UndoRedoAction, which is likely to be a good choice for them anyways. However, third-party plugins that use CreateComponent would be stuck with the Component initialization code. This shouldn't be too much of a problem for their own components since they define the initializers, but it would give them no ability to override Duality defined initializers, or those of other plugins they rely on. It is debatable that plugins should be able to overwrite the initializers defined by the Component's original creators though....

The other argument for removing them from CreateComponent is that initializing the Component is a side-effect that is not inherently obvious when using CreateComponent.

An alternative to removing them from CreateComponent is to change the constructor to take an optional boolean parameter that says whether the initializer should be ran. This would also fix the side-effect issue because users of CreateComponent would now be in control of whether or not the initializer is ran.

ilexp:
> Should the checks for initialization functions be removed from CreateComponent?

You made some good points.

Additionally, the `CreateComponent` UndoRedo action is actually used for both "regular" creation, as well as cloning (copy here) and parameterized creation (dragdrop a Material into the Scene), while the actual user editing setup EditorAction should only happen on regular creation, but not the other two usages. If the `CreateComponent` UndoRedo action should be the place for user editing setup code, it will need some modifications to distinguish between its various usages.

Won't manage to look into this in more detail today, but hopefully soon.

ilexp:
I've merged your changes and modified the `CreateComponent` undo / redo action accordingly. The only case where Components are now modified by user editing setup actions is when they are newly created - either as a dependency or directly. Components that are merely cloned or initialized using external parameters are not modified at all.

You can find the full list of changes [here](https://github.com/AdamsLair/duality/commit/3e904618e1854e801d58524357a18e4fa041fe95).

PR #292, opened 2016-02-18
Extended RenderGlyphs method

intercross21:
Now it accepts the ExtendedSet data container which allows to modify set of characters which will be drawn.
http://forum.adamslair.net/viewtopic.php?f=4&t=715#p4282

### DIFF
--- a/DualityEditor/Utility/Extensions/ExtMethodsDualityFont.cs
+++ b/DualityEditor/Utility/Extensions/ExtMethodsDualityFont.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Text;
 using System.Globalization;
 using System.Linq;
 using System.Drawing;
--- a/DualityEditor/Utility/Extensions/ExtMethodsDualityFont.cs
+++ b/DualityEditor/Utility/Extensions/ExtMethodsDualityFont.cs
@@ -28,8 +28,9 @@ namespace Duality.Editor
 
 		/// <summary>
 		/// Renders the <see cref="Duality.Resources.Font"/> based on its embedded TrueType representation.
+		/// <param name="extendedSet">Extended set of characters for renderning.</param>
 		/// </summary>
-		public static void RenderGlyphs(this DualityFont font, ExtendedCharSet extendedSet = null)
+		public static void RenderGlyphs(this DualityFont font, FontRenderGlyphCharSet extendedSet = null)
 		{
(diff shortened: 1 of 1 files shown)


DualityEditor/Utility/Extensions/ExtMethodsDualityFont.cs
```diff
@@ -113,12 +114,25 @@ public static void RenderGlyphs(this DualityFont font, FontFamily fontFamily)
...
-		private static void RenderGlyphs(DualityFont target, SysDrawFont internalFont)
+		private static void RenderGlyphs(DualityFont target, SysDrawFont internalFont, ExtendedCharSet extendedSet = null)
 		{
-			DualityFont.GlyphData[] glyphs = new DualityFont.GlyphData[DefaultChars.Length];
+            string allChars = DefaultChars;
+            string allCharBaseLineRef = CharBaseLineRef;
+            string allCharDescentRef = CharDescentRef;
+            string allCharBodyAscentRef = CharBodyAscentRef;
+
+            if(extendedSet != null)
```
ilexp: Code Style: Forgot to add a space between `if` and `(`.

DualityEditor/Utility/Extensions/ExtMethodsDualityFont.cs
```diff
@@ -113,12 +114,25 @@ public static void RenderGlyphs(this DualityFont font, FontFamily fontFamily)
...
+            {
+                allChars += extendedSet.Chars;                      //Maybe we should check if any char is duplicated
+                allCharBaseLineRef += extendedSet.CharBaseLine;     //And also check availbility of those characters in current font
```
ilexp: Code Style: Avoid comments at the end of a code line. Also add a space after `//` before the text begins.

ilexp:
Good addition! Some review in addition to inline comments:

- Please use tabs for indentation, rather than spaces.
- Replace public fields in `ExtendedCharSet` with public properties.
- Please provide XML documentation for every public property of `ExtendedCharSet`, as well as the class as a whole. The editor is chronically under-documented and I'm trying to improve here.
- Rename `ExtendedCharSet` to something like `FontRenderGlyphCharSet` to reflect its degree of specialization and provide context. It's in a very open namespace.

intercross21:
I changed few options in VS13 so now auto-correction should work properly.

ilexp:
Thanks! Lines 61, 110 and 120-123 still use spaces instead of tabs - I'll fix this when I do the merge, but please double-check next time :)

Issue #296, opened 2016-02-24
Create an Undo History Module


ilexp:
#### Summary

The Duality editor has full Undo / Redo support, but the only way to access it is via Ctrl + Z / Ctrl + Y shortcuts and a menu entry. It would be much cooler if there was a full "History" module showing the complete list of Undo / Redo steps and allowing to revert back to an old state with one click. For reference, maybe something along the lines of the [Photoshop](http://i.imgur.com/9TB4VOR.gif) or [Gimp](http://i.imgur.com/fYqZc7s.png) undo history.

#### Analysis

- This feature can easily be done in its own editor plugin, no changes to the core or editor are required for this.

PR #297, opened 2016-02-27
Undo History Module for issue #296

gb056:
Undo History Module for issue #296
Implemented as an editor plugin, but unfortunately had to expose UndoRedoManager.actionIndex and actionStack in the process.

This is my first contribution to an OS project so please feel free to give me all the criticism you have.


ilexp:
Hey gb056,

I gave your contribution a quick test and there are several points that would need to be addressed for this to be merged:

- Use tabs for indentation and spaces for alignment. Currently, there seems to be a complete mix of both.
- The background color of the docking control should match with the one of other editor modules, even if there are no items in the undo / redo queue.
- The Undo / Redo list being based on the Log View control was a good way to quickly produce a usable result, but I think it's not a good match overall:
  - Unlike in the log view, there is no concept of "selecting" an item in the Undo / Redo list. Being able to do so doesn't really serve a purpose.
  - There are no real columns in the undo / redo list, but displaying item numbers or other table-like data requires them.
  - There's not really a reason to use the monospaced font from the log view here and there should be an easy way to add icons and other info on log operations later.
- Instead of using a custom control for displaying the list, an often favored approach in the editor is to use the existing model / view `Aga.Controls.TreeView` control and implement a custom `NodeControl` if necessary. This also allows proper multi-column data. For reference, here are some spots in the editor that use this approach:
  - [Welcome Dialog](https://github.com/AdamsLair/duality/blob/master/DualityEditor/Forms/WelcomeDialog.cs)
  - [Package Manager Frontend](https://github.com/AdamsLair/duality/tree/master/DualityEditorPlugins/PackageManagerFrontend/Modules)
  - [Select Asset Importer Dialog](https://github.com/AdamsLair/duality/blob/master/DualityEditor/AssetManagement/SelectAssetImporterDialog.cs)
- The Undo / Redo list should perform a lazy update to avoid performance impact and flickering when rapidly modifying the undo / redo stack. You can observe this when dragging around an object in the scene view.
- Menu item and docking control shouldn't be named "Undo / Redo ..." as this indicates something along the lines of a user dialog that allows to perform some kind of a parameterized undo or redo operation. Instead, "Undo History" seems more straightforward to me. Also, the menu item should be in the View menu to match existing patterns.
- The icon has room for improvement. Maybe something like `table_refresh` from the [famfamfam icon set](http://www.famfamfam.com/lab/icons/silk/) could be used?

Sorry for the big list - I have to keep a close eye on the overall project quality, which includes not only code quality, but also usability and performance. Overall, I'd say what you've got is a solid work-in-progress piece, but not yet one that ready for being a part of the core repo.

gb056:
Great thanks for your feedback @ilexp.

The long list is not a problem at all, like I said I'm completely new to the OS game and figured this looked like a good place to just jump in and hack away as an experiment. Once it came together into something usable I figured I'd get your feedback.

I'll take a look at your list in detail later today and hopefully can have something you'd be happy to merge soon. Thanks for giving me the chance rather than just telling me to go away!

One last thing, you mentioned in the issue that "no changes to the core or editor are required for this." Is this still something you believe or were my changes to the UndoRedoManager necessary do you think?

ilexp:
Glad to hear! Let me know when you've finished up the next iteration.

> One last thing, you mentioned in the issue that "no changes to the core or editor are required for this." Is this still something you believe or were my changes to the UndoRedoManager necessary do you think?

No, I think you're right in this. There needs to be a way to enumerate the entire undo history and determine where the current version is inside it.

ilexp:
I have pushed an UndoRedoManager update where I've exposed the Undo / Redo stack in a minimal way. Please merge the current master branch back into yours before continuing development!

gb056:
Thanks, I've merged the changes locally now.

I have made a start on the improvements you suggested, but real life has got in the way a bit this week. I will hopefully continue to work on this during an evening or two this week though.

ilexp:
Closing this because there doesn't seem to be active development on this anymore.

PR #311, opened 2016-04-04
RenderTarget checks Texture size

mfep:
_A small usability addition_

When assigning textures to the render target, first it checks if the texture's size is set correctly (e.g. not 0). If not, it sends an error message to the user, and returns early.

This helps mostly beginner users, as it is not trivial that texture size has to be set before assigning.

### DIFF
--- a/Duality/Resources/RenderTarget.cs
+++ b/Duality/Resources/RenderTarget.cs
@@ -211,6 +211,14 @@ namespace Duality.Resources
 		/// </summary>
 		protected void SetupNativeRes()
 		{
+			foreach (var target in this.targets.Where(t => t != null).Res()) {
+				if (target.PixelWidth == 0 || target.PixelHeight == 0) {
+					Log.Core.WriteError("Error initializing RenderTarget {0}: Texture {1} has a dimension of 0 {2}",
+						this.FullName, target.FullName, Environment.NewLine);
+					return;
+				}
+			}
+
 			if (this.native == null) this.native = DualityApp.GraphicsBackend.CreateRenderTarget();
 
 			INativeTexture[] targets = this.targets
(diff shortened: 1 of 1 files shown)


Duality/Resources/RenderTarget.cs
```diff
@@ -211,6 +211,16 @@ protected void FreeNativeRes()
 		/// </summary>
 		protected void SetupNativeRes()
 		{
+			foreach (var target in this.targets.Where(t => t != null).Res())
+			{
+				if (target.PixelWidth == 0 || target.PixelHeight == 0)
+				{
+					Log.Core.WriteError("Error initializing RenderTarget {0}: Texture {1} has a dimension of 0 {2}",
+						this.FullName, target.FullName, Environment.NewLine);
```
ilexp: - No need to add a NewLine at the end of a log. The log system will do for you.
- Add single quotes (`'`) around `{0}` and `{1}` as they are referring to Resources.
- Instead of passing the `FullName` property, pass the Resource instance directly (`this` / `target`). The Log will display them using their full name anyway and you get double-click ability in the editor's LogView as a bonus.

ilexp:
Nice addition! :) I've added some comments in your code. Mostly regarding log message formatting.

mfep:
Thank you for guidance, the changes has been committed.

What I don't get is that my log message was based on the 226th line of the origninal code:

`Log.Core.WriteError("Error initializing RenderTarget {0}:{2}{1}", this.FullName, Log.Exception(e), Environment.NewLine);`

which does not seem to undergo the messaging standards you have just mentioned.

ilexp:
> What I don't get is that my log message was based on the 226th line of the origninal code:

> `Log.Core.WriteError("Error initializing RenderTarget {0}:{2}{1}", this.FullName, Log.Exception(e), Environment.NewLine);`

> which does not seem to undergo the messaging standards you have just mentioned.

Duality has a big code base and some of it existed before some of the guidelines applied today - you've used some old code as an example that, unfortunately, has not been updated since. 

The `NewLine` is actually applied in the middle of it though - despite counter-intuitively appearing at the end of the parameter list.

> Thank you for guidance, the changes has been committed.

Merged. Thanks for contributing! :)

Issue #285, opened 2016-02-14
Allow Custom Data Attachments in AssetInfo


ilexp:
Right now, the [AssetInfo](https://github.com/AdamsLair/duality/blob/master/Duality/Resources/AssetInfo.cs) class stores a hardcoded set of information. However, it would be very nice if AssetImporters could store additional information in there:

- The original file format this Resource was imported from, so it can be exported to a matching file.
- Custom import parameters, obtained by user interaction and re-used when available.
- It could actually replace the current `Font.EmbeddedTrueTypeFont` property, which would be a cleaner design anyway.
- ...others?

ilexp:
Design idea:

- Adjust the ResourcePropertyEditor to detect if there is additional data attached to the AssetInfo object and display a specialized AssetInfoPropertyEditor for this after the one for the actual resource.
  - This would allow to edit the settings that are used by the importer for re-import operations.
  - There should be a ReImport button at the end of this editor.
  - There should be a way to specify which items are displayed from within the importer.
  - Maybe the property editor could ask the preferred importer (as already specified in AssetInfo) for a settings object to display.
- This could actually solve multiple problems at once and make a very convenient way to configure Asset re-import operations after they've been imported for the first time.

ilexp:
**Progress:**

- Did a prototype implementation and discarded it again.

**Issue:**

- When defining editor-specific `AssetImporterConfig` classes that will be serialized as part of a `Resource`, there will be serialization errors when loading that `Resource` with the required editor plugin removed in a release.
- This will also restrict resources to only being loaded only after all editor plugins have been loaded as well.
- None of the above is acceptable.

**Solution:**

- Instead of letting an editor plugin define a class that holds an importer's configuration, split up the configuration of an import into N key-value pairs mapping from string (id) to object (value).
  - No knowledge about editor classes is required for serialization.
- The import / export environment can provide specialized API for accessing and modifying this configuration.
  - The API should handle all the casting and availability checking conveniently.
  - This API could also check newly assigned object values for their type to make sure that no editor plugin type is used, prompting the above issue on the user side.
- An `AssetInfoPropertyEditor` implementation would have to manually populate its property editors based on the currently available key-value pairs.
  - Dynamically adding and removing those pairs is not required, as it would not have been a requirement with config classes.
  - Editor visibility of individual pairs can be done using the naming convention that was introduced for shader variable hiding: Adding an underscore in the beginning flags a value as "private".

[...]

PR #315, opened 2016-04-20
Font Glyph Y Offsets, Tweaks

mika76:
Font changes to help with allowing bitmap fonts. 

I've tried to keep the changes as non-impacting as possible - ie - all previous configs should work - offsetY should be set to 0 by default and embedded fonts use it as such. In some cases I also had to change some checks where embedded font was checked explicitly because when only using the pixeldata there will be no embedded font.

@ilexp Please let me know what you think.


mika76:
I will do the actual Bitmap Font Importer as a separate project\package and put up a link here.

DualityEditor/Utility/Extensions/ExtMethodsDualityFont.cs
```diff
@@ -36,47 +36,53 @@ public static class ExtMethodsDualityFont
 		/// </summary>
 		public static void RenderGlyphs(this DualityFont font, FontRenderGlyphCharSet extendedSet)
 		{
-			if (font.EmbeddedTrueTypeFont == null) throw new InvalidOperationException("Can't render glyphs of a Duality Font without embedded vector Font information.");
```
mika76: This was changed purely because changing properties on a Font would make the preview fail since it's waiting on the EmbeddedFont. My thinking is with a bitmap font you'd want to recreate kernings (currently since they are always used if the arrays are set) only, while with a bitmap font, you might have changed size which should re-generate all the glyphs.

mika76:
You can see the importer at https://github.com/mika76/duality-bitmapfont-importer - I am not sure if it's working yet - I was developing it in the main Duality solution and only now have split it out - but will get onto that tomorrow...

SampleProjects/BasicMenu/BasicMenu.core.xml
```diff
@@ -4,21 +4,6 @@
         <name>BasicMenu.core</name>
```
mika76: Hmm, this was probably done by VS2015 - I'll undo the change...

ilexp:
Overall, I don't agree with your changes to `ExtMethodsDualityFont.cs`, `EditorBasePlugin.cs` and `FontContentPropertyEditor.cs`, but I think you did a good job implementing the Y offset.

Regarding the first point, I can see that what you're trying to do is enabling users to switch kerning on and off again, but I think that this should be done in a different way. In short, the `RenderGlyphs` method has a very specific purpose and should not be invoked if there are no glyphs to be rendered, thus should remain entirely unchanged. Instead, special case handling like this can be done in the property editor for Fonts and the reprocess on change snipped in the base plugin. 

I'll annotate the details inline in your code files shortly.

Duality/Resources/Font.cs
```diff
@@ -107,14 +107,19 @@ public struct GlyphData
 			/// The glyphs kerning samples to the right.
 			/// </summary>
 			public int[] KerningSamplesRight;
+			/// <summary>
+			/// The glyphs Y offset when rendering it.
+			/// </summary>
+			public int OffsetY;
```
ilexp: Please move this field up, so it's directly under `OffsetX`, just for reading flow reasons. Don't worry about binary compatibility, the `BinarySerializer` saves enough metadata to deal with this properly.

Duality/Resources/Font.cs
```diff
@@ -382,12 +387,15 @@ public void UpdateKerningData()
 
 				for (int i = 0; i < this.glyphs.Length; ++i)
 				{
-					PixelData glyphTemp = this.GetGlyphBitmap(this.glyphs[i].Glyph);
+					PixelData glyphTemp_ = this.GetGlyphBitmap(this.glyphs[i].Glyph);
+					PixelData glyphTemp = new PixelData(this.glyphs[i].Width, this.glyphs[i].Height + this.glyphs[i].OffsetY);
+					glyphTemp_.DrawOnto(glyphTemp, BlendMode.Solid, 0, this.glyphs[i].OffsetY);
```
ilexp: Creating another temporary glyph bitmap to draw it with an offset is really slow. For performance reasons, it would be much better to account for the Y offset math-wise in the following algorithm.

mika76:
Cool, thanks. It's much easier to work on this when you can actually see he code and comment instead of me asking a million questions.

The editor stuff I actually changed because I noticed the property editor preview actually failed when there was no embedded font and because of that all the properties became read only. I am not 100% on the changes myself but needed you to have a look and maybe give some pointers.

Duality/Resources/Font.cs
```diff
@@ -382,12 +387,15 @@ public void UpdateKerningData()
...
+					PixelData glyphTemp_ = this.GetGlyphBitmap(this.glyphs[i].Glyph);
+					PixelData glyphTemp = new PixelData(this.glyphs[i].Width, this.glyphs[i].Height + this.glyphs[i].OffsetY);
+					glyphTemp_.DrawOnto(glyphTemp, BlendMode.Solid, 0, this.glyphs[i].OffsetY);
+
 
 					this.glyphs[i].KerningSamplesLeft	= new int[kerningY.Length];
 					this.glyphs[i].KerningSamplesRight	= new int[kerningY.Length];
 
-					if (this.glyphs[i].Glyph != ' ')
+					if (this.glyphs[i].Glyph != ' ' && this.glyphs[i].Glyph != '\t' && !(this.glyphs[i].Height == 0 && this.glyphs[i].Width == 0))
```
ilexp: Change for readability reasons:
```
[...] && this.glyphs[i].Height > 0 && this.glyphs[i].Width > 0
```
Nice addition!

Duality/Resources/Font.cs
```diff
@@ -876,14 +890,15 @@ public int PickTextGlyph(string text, float x, float y)
...
-		private void ProcessTextAdv(string text, int index, out GlyphData glyphData, out Rect uvRect, out float glyphXAdv, out float glyphXOff)
+		private void ProcessTextAdv(string text, int index, out GlyphData glyphData, out Rect uvRect, out float glyphXAdv, out float glyphXOff, out float glyphYOff)
 		{
 			char glyph = text[index];
 			int charIndex = (int)glyph > this.charLookup.Length ? 0 : this.charLookup[(int)glyph];
 			this.texture.LookupAtlas(charIndex, out uvRect);
 
 			this.GetGlyphData(glyph, out glyphData);
 			glyphXOff = -glyphData.OffsetX;
+			glyphYOff = glyphData.OffsetY;
```
ilexp: X and Y offset should behave the same with regard to their axis orientation. Since we'll need to keep the negative X offset for compatibility reasons, please add a negative sign to the Y offset as well, so "offsets doing negative movement" is a consistent rule.

DualityEditor/Utility/Extensions/ExtMethodsDualityFont.cs
ilexp: See main comment in the PR :)

DualityEditorPlugins/EditorBase/EditorBasePlugin.cs
```diff
@@ -88,7 +88,7 @@ private void OnResourceModified(ContentRef<Resource> resRef)
 				if (resRef.IsLoaded)
 				{
 					Font fnt = resRef.As<Font>().Res;
-					if (fnt.GlyphsDirty && fnt.EmbeddedTrueTypeFont != null)
+					if (fnt.GlyphsDirty)
```
ilexp: This line should stay as it was before. We shouldn't invoke RenderGlyphs when there isn't actually anything to render. Instead, there could be an else case when `fnt.GlyphsDirty && fnt.EmbeddedTrueTypeFont == null` that only updates kerning specifically.

DualityEditorPlugins/EditorBase/PropertyEditors/Resources/FontContentPropertyEditor.cs
```diff
@@ -27,7 +27,7 @@ protected override void OnGetValue()
 			Font font = this.GetValue().FirstOrDefault() as Font;
 
 			bool lastCanRenderFont = this.canRenderFont;
-			this.canRenderFont = (font != null && font.EmbeddedTrueTypeFont != null);
+			this.canRenderFont = (font != null && font.GetGlyphBitmap(' ') != null);
```
ilexp: This line should remain unchanged as well, because we can, in fact, not render the font, if there is no embedded truetype font. Stating otherwise would be an error. 

(We might need to figure something out in order to allow the user to still access the kerning field non-readonly, but one thing at a time. That's probably for another PR.)

ilexp:
> The editor stuff I actually changed because I noticed the property editor preview actually failed when there was no embedded font

This is interesting, actually sounds like this might be a bug, or at least an oversight. The preview should work just fine in any case. What's the behavior you're experiencing, e.g. what does "preview failed" mean?

Also, I'm done with the inline comments. Wasn't too bad actually! Let me know when you've updated your PR and are in for the next review round.

Duality/Resources/Font.cs
mika76: I thought so too but did not understand the algorithm used and wanted something that works - also I thought since it's only really used when generating the font (generally in the importer) that it wasn't so risky. I'll have a look at the algorithm to see if I can update.

Duality/Resources/Font.cs
mika76: Interesting. I will definitely do this, but will have to test it some, since I'm not sure how it will impact the other calculations, especially kerning. Also I haven't seen this used in any of the generators for bitmap fonts.

DualityEditor/Utility/Extensions/ExtMethodsDualityFont.cs
mika76: Yes, this change was purely done to allow the property Kerning = true/false to recalculate. I can remove until we discuss kernings in more depth.

DualityEditorPlugins/EditorBase/PropertyEditors/Resources/FontContentPropertyEditor.cs
mika76: @ilexp  Can you explain what you mean by "Render" in this case? In the case of a bitmap font you will never have a bitmap font, but it can still be rendered (drawn to screen) no?

The problem I had with the property grid making all properties readonly was directly related to this  because without it running, the canRenderFont was set to false and therefore all properties were set to readonly.

ilexp: > @ilexp Can you explain what you mean by "Render" in this case? In the case of a bitmap font you will never have a bitmap font, but it can still be rendered (drawn to screen) no?

Ah, allright. That might have caused some confusion. When I speak of rendering a Font in this case, I mean re-rendering the .ttf vector data onto a set of glyph bitmaps. This has nothing to do with rendering a Font ingame, since all Fonts can be rendered in that sense if they contain any usable data.

> The problem I had with the property grid making all properties readonly was directly related to this because without it running, the canRenderFont was set to false and therefore all properties were set to readonly.

Yep, that is a problem for the kerning switch. We'll need to figure this one out separately.

Duality/Resources/Font.cs
mika76: Can you explain here a bit more. As far as I can see offset x and offset y are purely an  offset within the cell of the character block (ie the cell might be padded or the cells might be a constant width and height) so in all the cases I can't see how they would ever be negative - both x and y. Kernings are actually used for manipulating the x-axis to bring then closer but otherwise I can only see offset x and y being positive.

ilexp: This has nothing to do with kerning! All offsets can be both negative and positive, depending on the direction of the offset. They are not paddings, they are literally added to the position of a glyph when drawing it in a text block. 

What I was referring to in my initial comment is just that I think that their behavior should be similar in terms of which direction positive vs. negative values point to, and since the X offset is already defined with a minus sign in the above code, the Y offset should be too.

DualityEditorPlugins/EditorBase/PropertyEditors/Resources/FontContentPropertyEditor.cs
mika76: Ok - although this specific property canRenderFont is used in FontPropertyEditor.BeforeAutoCreateEditors to make all the properties of the font readonly. Can you suggest another way to fix this?

ilexp: > Can you suggest another way to fix this?

In the end, it will be fixed by me implementing issue #285 and updating Font accordingly, but I don't know when that will be yet. So, in the meantime, one way to deal with this would be to further customize the font "content" property editor so it makes a distinction between "TTF rendering" and "Duality data" properties, and only makes the TTF rendering ones readonly. 

This will of course be a bit hacky, but since this will be replaced at some point anyway, I think it should be fine. I can't really point you towards a specific solution, but I think it can be implemented in the existing [FontContentPropertyEditor](https://github.com/AdamsLair/duality/blob/master/DualityEditorPlugins/EditorBase/PropertyEditors/Resources/FontContentPropertyEditor.cs) provided you configure it using the special case in [FontPropertyEditor](https://github.com/AdamsLair/duality/blob/master/DualityEditorPlugins/EditorBase/PropertyEditors/Resources/FontPropertyEditor.cs).

Duality/Resources/Font.cs
mika76: Ok I think I misunderstood the offsets then - I'm going through the code now to see if the way I used offsetY is correct.

mika76:
@ilexp I've made most of the changes requested. It works, although the properties will still be readonly until we make some sort of change.

One thing I couldn't figure out was the UpdateKernings change - I can't seem to update the algorythm to work with blocks which are not equal heights (which is why I re-draw them), maybe you can help here, although this only happens on import of assets (and on changing kerning property I guess) so I'm not sure if the performance hit is a problem.

ilexp:
In lines 664, 735 and 884, you're subtracting the glyph Y offset while the X offset is added. This should be the same for both, basically the same argument as before in `ProcessTextAdv` - you've essentially moved the issue one level up :smile: X and Y offsets should behave the same with regard to their axis orientation.

As far as the readonly issue goes, this can be fixed by adjusting the FontPropertyEditor as described previously, but this another addition and not due for this PR.

The kerning update algorithm first picks a set of sample heights in an assumed constant-height glyph block and then, for each glyph, samples pixels on that height level to find how many pixels you'd need to travel from the edge until you find a non-transparent one. In order to make it work with different heights, you can still assume that constant-height glyph block, but offset each sample's Y coordinate using the Y offset value of the glyph (or the negative one, maybe) - and assume a transparent pixel when exceeding the available pixel data.

I can make the required adjustments myself if you don't mind your PR needing to wait a bit longer, since I'll have to allocate a bigger time slot for this.

mika76:
> In lines 664, 735 and 884, you're subtracting the glyph Y offset while the X offset is added. This should be the same for both, basically the same argument as before in ProcessTextAdv - you've essentially moved the issue one level up :smile: X and Y offsets should behave the same with regard to their axis orientation.

I know haha but when I did do it the way I got the following:
example

ilexp:
> I know haha but when I did do it the way I got the following:

To solve this, adjust your Font importer to provide the Y Offset values in the format that is used by Duality, i.e. flip them. I think BMFont might just use Y offsets that go the other way around.

(I know, this is kind of a rat's tail we're following here, but having OffsetX and Y behave differently is not a good idea and changing the orientation of the X offset is a breaking change, so.. this is probably the best we can do before the next major version change.)

mika76:
Ok, flipped them and if I reverse the offsets when I import then it seems it works.

mika76:
If you'd like to try it - get the importer from https://github.com/mika76/duality-bitmapfont-importer (just copy the cs file into duality - I can't create package until this pull request is accepted since it required Font to have OffsetY)

ilexp:
Allright, I'll consider this PR ready for merge and will do the kerning algorithm change in the process. Might take a few days, but you'll be notified when it's done.

ilexp:
Applied the required UpdateKerning changes and merged into master.

PR #320, opened 2016-04-29
Adding XAdvance

mika76:
Here's my attempt :smile: 

I've Added the AdvanceX property and moved the calculation of it to the font renderer instead of the update kernings method. I've also had to update your included fonts. Let me know what you think.

### DIFF
--- a/Duality/Resources/Font.cs
+++ b/Duality/Resources/Font.cs
@@ -171,10 +171,19 @@ namespace Duality.Resources
 		/// <summary>
 		/// [GET / SET] Specifies how the glyphs of this <see cref="Font"/> are rendered in a text.
 		/// </summary>
+		[EditorHintFlags(MemberFlags.AffectsOthers)]
 		public RenderMode GlyphRenderMode
 		{
 			get { return this.renderMode; }
-			set { this.renderMode = value; }
+			set
+			{
+				if (this.renderMode != value)
+				{
+					this.renderMode = value;
+					this.GenerateTexture(); // Filtering depends on pixel-grid alignment.
+					this.GenerateMaterial();
+				}
+			}
 		}
 		/// <summary>
 		/// [GET] The <see cref="Duality.Resources.Material"/> to use when rendering text of this Font.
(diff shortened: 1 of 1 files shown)


ilexp:
Quickly looked through the code, I think the implementation isn't quite there yet:

1. Spacing is a Font-wide setting that still needs to be applied as before. Users might want to adjust this, it is not replaced by the new explicit `AdvanceX` but needs to interact properly.
2. The Monospace setting can't go away either. Same as with spacing, needs to interact properly with the new `AdvanceX`, rather than being replaced.
3. Where does the "magic equation" `(internalFont.Size / 10.0f)` come from?

On a different note, updating the embedded Fonts is definitely required, good thinking!

mika76:
> 3 

It comes from the charSpacinng calculation in the current font implementation. As far as I could see bmfont included spacing in the xadvance? It all looks good when rendering. Monospace still seems to work too.

ilexp:
Actually, we need to do something about backwards compatibility in general, now that I think about it. Instead of updating the internal Font, there needs to be some preparation in `Font.OnLoaded` that checks if a sample set of glyphs has a non-zero `AdvanceX` value and auto-calculates it based on width and offset if not.

mika76:
Ah OK I'll have a look at that.

ilexp:
> It comes from the charSpacinng calculation in the current font implementation.

Ah. In that case, it should stay in the spacing variable so it's user-adjustable on a per-Font level.

> Ah OK I'll have a look at that.

Great, thanks :+1:

mika76:
Spacing was always calculated as far as I saw and was not user changeable...

mika76:
It's in the size setter.

ilexp:
> Spacing was always calculated as far as I saw and was not user changeable...

> It's in the size setter.

It's default is auto-calculated, but it's [adjustable by the user](https://github.com/AdamsLair/duality/blob/master/Duality/Resources/Font.cs#L207-L211) using the `CharSpacing` property in the object inspector.

mika76:
Ah of course. Ok I'll add it back. 

In pixel fonts it seems to me that spacing is included in the xadvance though, do you think I should negate it when importing?

ilexp:
> In pixel fonts it seems to me that spacing is included in the xadvance though, do you think I should negate it when importing?

I think for now you could set `CharSpacing` property to zero right after calling `SetGlyphData` in your importer. Since `Size` is never set for BMFonts, it won't jump back to the hardcoded default and you won't have to adjust the (correctly imported) x-advance values.

Issue #384, opened 2016-09-03
Provide ICmpTilemapRenderer API to Convert Tile Pos to Local Space Pos


ilexp:
#### Summary

The current `ICmpTilemapRenderer` API provides functionality to determine the tile position from a local space position, but not vice versa. This functionality should be added.

ilexp:
Closed, see merged PR.

Issue #298, opened 2016-02-27
Switch to Visual Studio 2015


ilexp:
#### Summary

Switch the main development repository of Duality to Visual Studio 2015 to gain access to new features like Analyzers. This should not affect the minimum required version to develop plugins with Duality, only developing Duality itself will be upgraded. The framework version should not be changed.

#### Analysis

- Visual Studio 2015 should provide access to the [Analyzers](http://www.dotnetcurry.com/visualstudio/1197/diagnostic-analyzers-visual-studio-2015) feature
- Code Analyzer availability is probably the main benefit of switching to VS 2015. This will allow to enforce coding style and other development guidelines and reduce the code review workload for new contributions.
- For Duality plugins, nothing should be changes. Users can switch versions as they prefer and compatibility should be a bigger concern. VS 2015 is still comparatively new and there is no good reason to enforce this as a minspec for framework users.

ilexp:
**Progress:**

- Reviewed the available Analyzer ecosystem. A lot of them are big packages of rules where a large percentage isn't really usable for Duality or has never been a problem in code reviews, yet slows down compilation a lot. Some other rules that would have been expected or preferred do not exist.
- Right now, a custom set of Duality-specific Analyzers seems preferable. That way, specific rules that are derived from code review experiences can be written and maintained one at a time.

ilexp:
**Notes:**

- Apparently, the default way to deploy custom code analyzers is via NuGet package. Since there is no dependency between the analyzers and Duality itself, it would be a viable option to completely split the two and have all the Analyzers in a different solution file, or maybe even repository.

ilexp:
**Notes:**

- Looking at the [complexity](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/fdb043f5ec9601649a992a43f0b76b62bfcfe4f4/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1027UseTabsCorrectly.cs) of [existing code style analzers](https://github.com/DotNetAnalyzers/StyleCopAnalyzers), it seems like a bad idea to custom-write all required analyzers. Reconsider using existing style analzers.

PR #392, opened 2016-09-14
#384 Tile position to local position in ICmpTilemapRenderer

mfep:
The solution for issue #384.

### DIFF
--- a/DualityPlugins/Tilemaps/Tilemaps/Components/ICmpTilemapRenderer.cs
+++ b/DualityPlugins/Tilemaps/Tilemaps/Components/ICmpTilemapRenderer.cs
@@ -47,6 +47,13 @@ namespace Duality.Plugins.Tilemaps
 		/// <returns></returns>
 		Point2 GetTileAtLocalPos(Vector2 localPos, TilePickMode pickMode);
 		/// <summary>
+		/// Gets the local position of the specified tile. The function does not check if the point is a valid tile position.
+		/// </summary>
+		/// <param name="tile">The position of the tile.</param>
+		/// <param name="alignment">Determines the point of the tile that is returned.</param>
+		/// <returns></returns>
+		Vector2 GetLocalPosAtTile(Point2 tile, Alignment alignment = Alignment.Center);
+		/// <summary>
 		/// Determines the generated depth offset for the tile at the specified tile coordinates.
 		/// This also inclues the renderers overall depth offset.
 		/// </summary>
--- a/DualityPlugins/Tilemaps/Tilemaps/Components/ICmpTilemapRenderer.cs
+++ b/DualityPlugins/Tilemaps/Tilemaps/Components/ICmpTilemapRenderer.cs
@@ -27,6 +27,10 @@ namespace Duality.Plugins.Tilemaps
 		/// </summary>
 		Rect LocalTilemapRect { get; }
 		/// <summary>
+		/// [GET] Returns the size of the bounding rectange of one tile.
+		/// </summary>
+		Vector2 LocalTileSize { get; }
+		/// <summary>
 		/// [GET / SET] A color by which the rendered <see cref="Tilemap"/> is tinted.
 		/// </summary>
 		ColorRgba ColorTint { get; set; }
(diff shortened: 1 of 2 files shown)


mfep:
So appveyor seems to not like the `?.` operator. I suppose I have to solve the nullreference checks the other way, right?

ilexp:
> So appveyor seems to not like the ?. operator. I suppose I have to solve the nullreference checks the other way, right?

The entire Duality repository uses C# 5.0, while the operator you're using was introduced in, I believe, C# 6.0 or later.

mfep:
Okay, it is done the exception catching way now. By the way, would it be a major enterprise to switch to C# 6.0? It has some features that make one's life easier, for example this *Elvis* operator above, or the `$` syntax for `String.Format`

ilexp:
> By the way, would it be a major enterprise to switch to C# 6.0?

No, and it will definitely happen - but not right now. It's one of those things where I like to lag behind a bit for compatibility reasons. There's no rush, and some users or contributors might still be using C# 5.0, and I don't want to force people to update. That said, issue #298 will likely require and trigger an update to a newer C# and build system version.

Will do some code review things now, brb.

DualityPlugins/Tilemaps/Tilemaps/Components/TilemapRenderer.cs
```diff
@@ -201,6 +201,58 @@ public Point2 GetTileAtLocalPos(Vector2 localPos, TilePickMode pickMode)
...
 		}
 		/// <summary>
+		/// Gets the local position of the specified tile. The function does not check if the point is a valid tile position.
+		/// </summary>
+		/// <param name="tile">The position of the tile.</param>
+		/// <param name="alignment">Determines the point of the tile that is returned.</param>
+		/// <returns></returns>
+		public Vector2 GetLocalPosAtTile(Point2 tile, Alignment alignment = Alignment.Center)
+		{
+			try
```
ilexp: Do not use try-catch to avoid `NullReferenceExceptions` - it's really, really inefficient compared to the alternative of just checking for null and avoiding access.

You should be able to get away with two null checks: `this.ActiveTilemap` and the `tileset` you get by accessing `this.ActiveTilemap.Tileset.Res`. This will also allow you to fall back to the default tile size. Suggestion: Mirror [this](https://github.com/AdamsLair/duality/blob/master/DualityPlugins/Tilemaps/Tilemaps/Components/TilemapRenderer.cs#L232-L236).

DualityPlugins/Tilemaps/Tilemaps/Components/TilemapRenderer.cs
```diff
@@ -201,6 +201,58 @@ public Point2 GetTileAtLocalPos(Vector2 localPos, TilePickMode pickMode)
...
+			{
+				Vector2 tileSize = ActiveTilemap.Tileset.Res.TileSize;
```
ilexp: When accessing local members, use `this.Xy` by code style convention.

DualityPlugins/Tilemaps/Tilemaps/Components/ICmpTilemapRenderer.cs
```diff
@@ -47,6 +47,13 @@ public interface ICmpTilemapRenderer : ICmpRenderer
 		/// <returns></returns>
 		Point2 GetTileAtLocalPos(Vector2 localPos, TilePickMode pickMode);
 		/// <summary>
+		/// Gets the local position of the specified tile. The function does not check if the point is a valid tile position.
+		/// </summary>
+		/// <param name="tile">The position of the tile.</param>
+		/// <param name="alignment">Determines the point of the tile that is returned.</param>
+		/// <returns></returns>
+		Vector2 GetLocalPosAtTile(Point2 tile, Alignment alignment = Alignment.Center);
```
ilexp: I think the Alignment argument is a nice idea, but it shouldn't be part of the interface specification, as this will force all implementations to copy-paste the same logic. For now, I'd suggest to simply remove it because it's quite simple to calculate yourself if you have the tile size.

Instead, the position that should be returned is the upper left one.

ilexp:
Allright, done with requesting changes. Nice PR overall :+1:

ilexp:
Oh, another one! It would be really convenient if `ICmpTilemapRenderer` exposed a `Vector2 LocalTileSize { get; }` property so things like alignment calculation are easier to do with less code.

DualityPlugins/Tilemaps/Tilemaps/Components/ICmpTilemapRenderer.cs
mfep: Okay, if I'm getting it right, if there was an implementation for, e.g a hexagonal tilemap renderer, that alignment wouldn't have a lot of sense.

ilexp:
Merged. Thanks! :+1:

Issue #371, opened 2016-08-16
Remove Camera Rotation in CamView


ilexp:
#### Summary

When holding the right mouse button, the CamView will rotate its camera. In the five years of developing Duality and using it to build various games and prototypes, I have never once used this, nor seen anyone using this. At the same time, rotation is confusing and difficult to undo. Consider removing CamView camera rotation entirely.

ChristianGreiner:
I wouldn't remove this. Why don't you create a toggle button for this feature?

ilexp:
> Why don't you create a toggle button for this feature?

Less code is often better with regard to maintainability, and overall trying to keep the codebase slim.

> I wouldn't remove this. 

Can you think of a scenario where you or someone else would miss it? Genuine question, will gladly keep this feature if it turns out to be useful.

ChristianGreiner:
> Can you think of a scenario where you or someone else would miss it? Genuine question, will gladly keep this feature if it turns out to be useful.

Mhm ok, you are right.  :see_no_evil:

ilexp:
> Mhm ok, you are right.  :see_no_evil:

Allright :smile: 

I'll leave this open for a while, so others get the chance to veto in, should there actually be viable uses.

mfep:
I absolutely agree with removing this.

ilexp:
It's up for grabs and, I assume, more of a low-hanging fruit, so if anyone wants to try their luck, go ahead. Basically just finding the right code and deleting it. If no one else does it, I'll remove it eventually.

mfep:
Right, I'll do it then

mfep:
A new discovery I just made: while pressing `space`, cursor movement determines the actual rotation of the camera, instead of the angular speed. This is much more controllable, than the *without-space* behavior, and less likely to accidentally fire it. I suggest to keep that. Which made this feature more usable, was to quantize the rotation to a fixed step of increment.

ilexp:
I'd be in favor of removing rotation entirely. Freeing up the right cursor key completely could be worth it.

ilexp:
PR merged. Closing this.

PR #396, opened 2016-09-17
Removed camera rotation in CameraView

mfep:
as discussed in issue #371


ilexp:
Nice and simple. Merged.

Issue #382, opened 2016-08-29
Can't delete a Component if the GameObject misses one of its RequiredComponent


ChatPion:
#### Summary

Basically what title says

#### How to reproduce [Bugs]

- Create a Component A
- Add Component A to a GameObject
- Create Component B
- Add B as RequiredComponent for A
- You can't delete A as long as you don't add B to the GameObject, 
EDIT : you must do it in the object inspector

#### Workaround [Bugs]

- Add missing RequiredComponent to the GameObject
- EDIT : Right-click + delete in the scene view works.

SirePi:
It appears this is no longer the case? I tried following the instructions but the Component A is being deleted without problems

ilexp:
Since this issue wasn't addressed in the mean time, I suspect that the reproduction steps might miss a step. 

@ChatPion Can you re-check if the steps above reproduce the error? Anything unmentioned to be aware of?

ChatPion:
This is strange.

Here is an ultra minimalistic project where I litterally followed every step.
[Duality Application.zip](https://github.com/AdamsLair/duality/files/503488/Duality.Application.zip)

I get this weird silly message : 
![2016-09-30_19-29-28](https://cloud.githubusercontent.com/assets/3697923/19000754/a08c6366-8744-11e6-9df1-d37bc4deb40c.png)

ilexp:
Uhm.. what you're doing there is specifying a cyclic requirement. This can't work out. A component shouldn't require itself.

Can you reproduce the same behavior without a cyclic reference?

mfep:
This issue happens, when the component dependencies change with recompiling the GamePlugin, @ChatPion means *create a .cs file, fill it, and recompile* by *Create*.
I propose that every plugin-reload should come with an iteration over the components if everyone's dependency is met, and add the missing components. Same for scene loading.

mfep:
Solution is on the way. Pull request incoming (most probably tomorrow).

ilexp:
> I propose that every plugin-reload should come with an iteration over the components if everyone's dependency is met, and add the missing components. Same for scene loading.

I don't think this is a good idea. Suppose you have a scene with 2000 objects, each having, say, 5 Components, you're going to check dependencies for 10000 Components every time you reload the plugin, only to catch that one corner case? 

[...]

PR #403, opened 2016-10-01
Can't delete a Component if RequiredComponent missing

mfep:
Fix of issue #382 
There are multiple ways to solve this. The one 'below' seemed the best, because:

* it's simple
* it's nonsene to check wheter the component we're going to remove has itself as a dependency

enjoy 🍕


ilexp:
This may be a bit silly to ask in this PR, but can you move your fix directly into the `IsComponentRequirementMet` implementation? I think it's general-purpose enough for that, and there might be other code relying on the same functionality that needs the fix as well.

mfep:
I'd argue that. Returning `true` in `IsComponentRequirementMet` when it's obviously not met feels wrong. It's used only in 2 cases in the code, shouldn't be too much hassle to correct them both. Your call in the end of course.

ilexp:
Hmm.. you may have a point. Merged.

Issue #405, opened 2016-10-02
Object inspector crashes when previewing zero sized Texture


ChatPion:
### Summary

What title says

### How to reproduce [Bugs]

- Create an empty pixmap
- Right-click on it -> create a texture
- Object inspector crashes, because it tries to resize a zero sized texture

A PR is incoming.

PR #406, opened 2016-10-02
Fixes crash after creating a texture from an empty pixmap

ChatPion:
Fixes [#405
](https://github.com/AdamsLair/duality/issues/405)


ilexp:
Nice! Thanks for your quick bug report and fix 👍 

Requesting one change before merging though: Please replace the data length check with a width or height zero check, because a data block can always be larger than it needs to be for a given width or height. 

(It's not currently the case, but trying to keep the code semantics clean for that elusive day in the future where we might implement pooling techniques.)

ilexp:
Thanks! Merged.

Issue #377, opened 2016-08-20
Remember the previous opened scene and switch to it when exiting the sandbox-mode


ChristianGreiner:
#### Summary
You want to switch to the the next scene (level) by code and you are currently in the sandbox mode, you have to switch to the previous scene (level) in the editor, because the editor stays at the new loaded scene even if you exist the sandbox-mode.

#### Analysis

- The editor have to remember the previous opened scenes.

ilexp:
Yep, that can be annoying at times. This should be improved.

ChristianGreiner:
Is the [Sandbox](https://github.com/AdamsLair/duality/blob/master/DualityEditor/Sandbox.cs) the right place to keep track of the previous scene?

So you could restore the previous scene, when the sandbox is stopped.

ilexp:
Yep! The way I see it, you'd temporarily store a ContentRef to the Scene you started the sandbox with, and when stopping, you'd return to that one [after resetting all Scenes](https://github.com/AdamsLair/duality/blob/9d3211a1b3c4fe5d7fd01c7da06359cac2cc2333/DualityEditor/Sandbox.cs#L157-L169).

ChristianGreiner:
We also should introduce a way to restore the stared scene when the engine is terminated in the editor. 
At the moment, the started scene gets removed, so you have to open the scene again.

ilexp:
Shoud be fixed with [this commit](https://github.com/AdamsLair/duality/commit/11f203db1dcb4ba558056d7b2bc7ece2eab6d0d3). No binary release yet though.

PR #407, opened 2016-10-02
Sandbox Returns to Started Scene After Stopping

ChristianGreiner:
Added a temporary scene in the sandbox.cs, so the sandbox restores the started scene.
Issue #377


ilexp:
Thanks for the PR! Looks good so far. Merged.

PR #409, opened 2016-10-09
Changed 'TimeFreeze' into boolean

ChristianGreiner:
In this pull request the **timefreeze** member (int) is now a boolean.
On top of that you can check with a public property (get), if the time is now frozen or not.

### DIFF
--- a/Duality/Time.cs
+++ b/Duality/Time.cs
@@ -100,7 +100,7 @@ namespace Duality
 		}        //	G
 
 		/// <summary>
-		/// [Get] Whether or not the game time is frozen.
+		/// [GET] Whether or not the game time is frozen.
 		/// </summary>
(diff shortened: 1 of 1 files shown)


ilexp:
I'll have to reject that one. Sorry!

Time freeze is designed for push / pop behavior, i.e. when you call freeze five times, you'll have to call resume five times for the simulation to actually resume. That way, each freeze site doesn't have to have any knowledge about any other and site A can't accidentally un-freeze before site B is ready as well. This is really useful for editor usage, where (for example) a CamView tool might suspend simulation while performing a dragdrop operation, while at the same time any other part of the editor or a plugin could do so as well.

Concluding this, time freeze is actually intended to be an `int` for internal counting purposes.

PR #422, opened 2016-10-24
Fix #419

### LINKED ISSUE #419: Object Visibility Filter in Camera View Remembers Filtering for Non-Existent / Invalid Component Types
_New Issue Text_

As described below, it is possible to select a Component for filtering in the Camera View, then delete the Component type entirely, rebuild the plugin and still have the (now invalid) filtering active, not displaying any object. This is likely also a problem even for still existing types after plugin reload, since the stored `Type` instances will no longer be valid.

------
_Original Issue Text:_

Right now, to hide a gameobject, we have to select all of its components. This is annoying because if I want to hide all my SpriteRenderers, I'll have to hide Transforms too for example.

How to fix this ? 
- easiest option : hide a gameoject if one of its component is hidden
- hardest but maybe best option : make it so we can hide components separately. For example, being able to hide SpriteRenderers while keeping all RigidbodyRenderer.

By the way, is it even useful to see in the list components that doesn't render anything ?

ChatPion:
Added checked items to the list even if not present in the scene.

### DIFF
--- a/DualityEditorPlugins/CamView/Modules/CamView.cs
+++ b/DualityEditorPlugins/CamView/Modules/CamView.cs
@@ -558,7 +558,7 @@ namespace Duality.Editor.Plugins.CamView
 
 			typesWithCount = (
 				from typeInfo in typesWithCount
-				where typeInfo.Count > 0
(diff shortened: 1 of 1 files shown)


ilexp:
Nice and simple. Merged. Thanks!

Issue #374, opened 2016-08-19
Add a Colorpicker in the ColorChoose-Dialoge


ChristianGreiner:
It would be helpful if you could pick a color from the current visible scene,

So we need a colorpicker in the "Choose a new color"-Dialoge for example.

ilexp:
Done by @SirePi. Triggered a release, check for package updates in about 30 minutes.

ChristianGreiner:
Sorry, when I'm digging out this old issue, but I like to add something:
I didn't mention that I suggested to add the colorpicker-button also in the "Choose a new color"-Dialoge.

![knancwi](https://cloud.githubusercontent.com/assets/6233308/20226696/64e1a562-a849-11e6-975d-f55cb5d7d97e.png)
 
So you can tweak your chosen color directly with the color wheel instead of picking up the color first, open the "Choose a new color"-Dialoge and then adjust the color.

ilexp:
Shouldn't be too big of an issue to add this. As the color dialog is defined in `AdamsLair.WinForms`, but color picking is implemented in the Duality editor itself, we have a problem. Solution approaches:

- Let the editor inject an additional button into the existing layout of the color picker dialog. Hacky, because it would need to be done every time a color dialog is instantiated. Not very nice.
- Move color picking to `AdamsLair.WinForms`. Investigate the implications of this.

Currently in favor of option two.

PR #424, opened 2016-10-26
Colorpick

SirePi:
Implementing https://github.com/AdamsLair/duality/issues/374

It uses a global low level mouse hook to listen to mouse Move events when the color pick button is pressed, and stops listening once the Left or Right mouse button is clicked; **Left** mouse confirms the change, **Right** mouse reverts to the last value of the editor.
This click (a single one) is filtered out, so it's safe to click on buttons or other UI elements, even of other windows, without issuing the relative command.

The color can be picked everywhere on the screen, even in a multi-monitor setup.

Also fixes a (bug?) where buttons were set to Hot state whenever the IColorDataPropertyEditor was set to Focused, independently if the button was actually hovered or not.


ilexp:
Merged. Refactored the picking code a little so it uses its own, self-contained class and can be re-used across the editor when needed.

PR #442, opened 2016-11-26
Better management of UserData changes

SirePi:
This change in NativeWindow.OnUserDataChanged performs a check to verify if there is really the need to update/recreate the window, based on its current and the new settings, and allows to switch the mouse cursor visibility at runtime.


ilexp:
You appear to have accidentally replaced the entire document's tabs with spaces. Please fix :)

SirePi:
Forgot to comment that, in addition, it makes possible to have a Resizeable window again, after the UserData has changed during runtime. Previously it was not the case.

ilexp:
Merged. 

There were a lot of non-changes in this commit, whitespace and alignment in unrelated places. Please try to stay closer to the PR again next time!

Issue #448, opened 2016-12-08
Adjust the Time Class to Favor per-Second Metrics Over per-Frame Metrics


ilexp:
### Summary

An object moves "100 units per frame" is harder to grasp than "100 unity per second", yet the reliance on `Time.TimeMult` reinforces per-frame metrics. To encourage users to do per-second metrics, make per-second calculations "first class citizens" in [`Time` API](https://github.com/AdamsLair/duality/blob/2f648f4874b8b3f82aff059324ab5e3382c62500/Source/Core/Duality/Time.cs).

### Analysis

- Remove `LastDelta`.
- Add a new `DeltaTime` instead that provides the time-scaled number of seconds that passed since last frame as a float. It can be a shorthand for `timeMult * SPFMult` internally.
- Rename `FPSMult`, `MsPFMult` and `SPFMult` to something more reasonable, even if it ends up being more verbose. With the new `DeltaTime` property, they're usually not needed anyway, and if they are it should be clear what they stand for.
  - New names could be "FramesPerSecond", "MillisecondsPerFrame" and "SecondsPerFrame".

ilexp:
Done by @mfep

PR #449, opened 2016-12-09
Changes in Time.cs according to issue #448

mfep:
Changes in Time.cs according to issue #448

### DIFF
--- a/Source/Core/Duality/Resources/Scene.cs
+++ b/Source/Core/Duality/Resources/Scene.cs
@@ -25,7 +25,7 @@ namespace Duality.Resources
 	[EditorHintImage(CoreResNames.ImageScene)]
 	public sealed class Scene : Resource
 	{
-		private const float PhysicsAccStart = Time.MsPFMult;
+		private const float PhysicsAccStart = Time.MillisecondsPerFrame;
 
 
 		private static World               physicsWorld      = new World(Vector2.Zero);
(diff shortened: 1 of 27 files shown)


mfep:
ehh...I don't know what I'm doing :(

Source/Core/Duality/Resources/Scene.cs
```diff
@@ -491,9 +491,9 @@ internal void Update()
 
 			// Update low fps physics state
 			if (!physicsLowFps)
-				physicsLowFps = Time.LastDelta > Time.MsPFMult && physTime > Time.LastDelta * 0.85f;
```
mfep: what do you think about this?

Source/Core/Duality/Resources/Scene.cs
```diff
@@ -491,9 +491,9 @@ internal void Update()
...
+				physicsLowFps = Time.DeltaTime > Time.MillisecondsPerFrame && physTime > Time.DeltaTime * 0.85f;
```
mfep: Do you think if changing `LastDelta` to `DeltaTime` is sufficient in the `physicsLowFps` test?

ilexp: Nope, because DeltaTime is affected by time scale. I'll merge it and replace it with something more reliable.

PR #466, opened 2016-12-27
Fixed a bug in Tile.cs ...

mfep:
... which has thrown an exception when UpdateAutoTileCon was called with null updateMask.


ilexp:
Thanks! Merged :)

Issue #453, opened 2016-12-11
Wrong AnimFirstFrame in AnimSpriteRenderer depending on the frame count


ChatPion:
### Summary

If the AnimFrameCount is one or more, the AnimFirstFrame is normal.

However, if it's equal to zero, then everything is offset by one : 
AnimFirstFrame               : 0 1 2 3 ...
Actual` index in spritesheet : 0 0 1 2 ...

ilexp:
I'm not 100% sure what you mean by "everything is offset by one". Can you provide a short source code excerpt and comment on the expected vs. the actual values of affected properties and methods?

ChatPion:
From what I experienced,
AnimFrameCount = 0
![paintdotnet_2016-12-12_20-22-41](https://cloud.githubusercontent.com/assets/3697923/21113173/cf407498-c0a8-11e6-9755-1f2e9660f0d7.png)

AnimFrameCount >= 1
![paintdotnet_2016-12-12_20-22-14](https://cloud.githubusercontent.com/assets/3697923/21113177/d0dbf660-c0a8-11e6-936a-cfb82b33c88e.png)

ilexp:
That makes it a bit clearer, thanks! I'll leave this open and up for grabs for now in case someone wants to tackle it. Doesn't seem that critical and could be a good case for new contributors.

### Workaround

Never assign less than one to `AnimFrameCount` - it doesn't usually make much sense anyway.

IanTeo:
Hi, is it ok for me to try to do this?

ilexp:
Yep 👍

IanTeo:
Are there any automated tests that I can run for this project after I have made the changes?
I looked around the contributing guide and could not figure out how to run the tests.
Could someone point me in the right direction? Sorry for the inconvenience.

ChristianGreiner:
I think you can use this nunit project:
https://github.com/AdamsLair/duality/blob/master/Test

ilexp:
Alternatively, you can select one of the test project as startup project in Visual Studio and it should launch the NUnit project @ChristianGreiner mentioned.

ilexp:
Fixed by @IanTeo

PR #481, opened 2017-01-19
Update AnimFrameCount default to 1

IanTeo:
## Changes

- Set default of animFrameCount to 1
- Prevent user from setting animFrameCount below 1
- Update EditorHintRange

Issue #453


ilexp:
Thanks!

If I could issue one wish, maybe add a few lines on your reasoning next time, for example: Why did you choose to prevent `animFrameCount` to be zero, as opposed to adjusting the internal methods to handle zero as a special case? It's clear to me now, but I had to dive into the source code first in order to find out. No big deal, just thought I'd mention :)

In any case, congrats to your first PR! Merged.

Issue #488, opened 2017-01-24
In the Editor, Initialize Textures With a Reasonable Size


ilexp:
### Summary

When creating a new Texture resource "from scratch" in the editor (without source Pixmap), it should be initialized with a reasonable default size, like `128x128`. 

### Analysis

- Initialize empty textures with a default size of `128x128`.
- Otherwise, it's quite easy to trip into errors thrown by other parts of the core, complaining about the zero-size texture.
- The setup shouldn't be done in the default constructor, but in an `EditorAction` with a setup context, like [this one](https://github.com/AdamsLair/duality/blob/066bfc5753dfabf895b391daa286787d28bf5185/Source/Plugins/EditorBase/EditorActions/SetupRigidBodyForEditing.cs). Instantiating new textures via source code should not imply allocating some arbitrary default space.
- Also, consider raising the editor hint min value for the [texture size property](https://github.com/AdamsLair/duality/blob/066bfc5753dfabf895b391daa286787d28bf5185/Source/Core/Duality/Resources/Texture.cs#L170) to one.

ilexp:
Done by @Barsonax

PR #490, opened 2017-01-24
master_488: When creating a new empty texture it is now initialized t…

Barsonax:
Textures are not initialized to 128x128 pixels instead of 0x0 pixels. Also changed the editor hint for the size to be between 1 and int.max.

### DIFF
--- a/Source/Plugins/EditorBase/EditorActions/SetupTextureForEditing.cs
+++ b/Source/Plugins/EditorBase/EditorActions/SetupTextureForEditing.cs
@@ -1,0 +1,22 @@
+﻿using Duality.Resources;
+
+namespace Duality.Editor.Plugins.Base.EditorActions
+{
+	public class SetupTextureForEditing : EditorSingleAction<Texture>
+	{
+		public override bool CanPerformOn(Texture texture)
+		{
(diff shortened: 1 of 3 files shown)


Source/Plugins/EditorBase/EditorActions/SetupTextureForEditing.cs
```diff
@@ -0,0 +1,22 @@
...
+	public class SetupTextureForEditing : EditorSingleAction<Texture>
+	{
+		public override bool CanPerformOn(Texture texture)
+		{
+			return base.CanPerformOn(texture) && texture.PixelWidth == 0 && texture.PixelHeight == 0;
+		}
+
+		public override void Perform(Texture texture)
+		{
+			texture.Size = new Vector2(128, 128);
```
ilexp: Missing Texture.ReloadData() after assigning the new size. Otherwise, the internal texture will still be uninitialized.

Barsonax: Added Texture.ReloadData()

ilexp:
Looks good so far. I'll test this locally in the next few days to make sure this doesn't break stuff. (Don't expect it though.)

ilexp:
Note: Addresses issue #488

Issue #492, opened 2017-02-02
Open sub folder when mouse hovering the root folder in the scene view


ChristianGreiner:
### Summary

In my opinion it's quite annoying when you want to import an asset from the windows explorer into the scene view (or an already imported asset) and the folder that you are hovering doesn't open automatically after a few seconds. 

![](http://i.imgur.com/r5mf7Dm.gif)

So every time you have to release the mouse button to know which folders are stored in the root folder...

I know you could also open the folder directly like this:
![](http://i.imgur.com/IvQpLWB.gif)
... but in my opinion it's not very user friendly.

What do u think?

ilexp:
Looks like a straight-forward usability improvement to me 👍 

A quick fix / test can probably be implemented in Project View and Scene View separately first. A proper longer-term solution would be to extend the behavior of [TreeViewAdv](https://github.com/AdamsLair/treeviewadv) directly.

ilexp:
Fixed by @nhold and awaiting binary release.

PR #508, opened 2017-03-07
Auto-Expand sub-folder when dragging over it. Issue #492

nhold:
### Auto-Expand for TreeViewAdv. Issue #492.

I went with the approach to extend TreeViewAdv and used the new class in both ProjectFolderView and SceneView. I tried to keep the changes as minimal as possible although the solution and csproj seem to have run away from a simple add existing file, let me know if I should remove them.

A few questions\comments:

1. Do you want to keep consistency of the type held by the SceneView\ProjectFolderView? I have currently kept the type as TreeViewAdv and just newed it to the new type (AutoExpandTreeView) which has a default of 1000 ms delay.

2. Is there a place to provide user\editor preferences, I would suggest putting the delay as an option for users. 

3. Have fun reviewing <3

4. Should I add automated testing?

Manual Tests:

test
test2

### DIFF
--- a/Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
+++ b/Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
@@ -1,0 +1,68 @@
+using Aga.Controls.Tree;
+using System;
+using System.Drawing;
+using System.Windows.Forms;
+
+namespace Duality.Editor.Controls
+{
+	public class AutoExpandTreeView : TreeViewAdv
+	{
+		public int DragOverExpandDelay
+		{
+			get;
+			set;
+		}
+
+		private TreeNodeAdv autoExpandNode = null;
+
+		private Timer timer = new Timer();
+
+		private void StartTimer()
+		{
+			if (DragOverExpandDelay > 0)
+			{
+				timer.Stop();
+				timer.Interval = DragOverExpandDelay;
+				timer.Tick += OnTimerTick;
+				timer.Start();
+			}
+		}
+
+		private void StopTimer()
+		{
+			if (DragOverExpandDelay > 0)
+			{
+				if (autoExpandNode != null)
+				{
+					autoExpandNode.Expand();
+				}
+			}
+
+			timer.Stop();
+		}
+
+		private void OnTimerTick(object sender, EventArgs e)
+		{
+			StopTimer();
+		}
+
+		protected override void OnDragOver(DragEventArgs drgevent)
+		{
+			base.OnDragOver(drgevent);
+
+			if (DragOverExpandDelay > 0)
+			{
+
+				Point p = PointToClient(new Point(drgevent.X, drgevent.Y));
+				var node = GetNodeAt(p);
+
+				if (autoExpandNode == null || autoExpandNode != node)
+				{
+					autoExpandNode = node;
+					StartTimer();
+				}
+			}
+
+		}
+	}
+}
--- a/Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
+++ b/Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
@@ -5,6 +5,11 @@ using System.Windows.Forms;
 
 namespace Duality.Editor.Controls
 {
+	/// <summary>
+	/// An extended TreeViewAdv.
+	/// Will automatically expand nodes that are being dragged over with another node.
+	/// A value of 0 in the constructor or setting the property AutoExpandDelay to 0 will disable this behaviour.
+	/// </summary>
 	public class AutoExpandTreeView : TreeViewAdv
 	{
 		private TreeNodeAdv autoExpandNode = null;
@@ -33,9 +38,7 @@ namespace Duality.Editor.Controls
 		{
 			if (AutoExpandDelay > 0)
 			{
-				autoExpandTimer.Stop();
 				autoExpandTimer.Interval = AutoExpandDelay;
-				
 				autoExpandTimer.Start();
 			}
 		}
@@ -44,10 +47,7 @@ namespace Duality.Editor.Controls
 		{
 			if (AutoExpandDelay > 0)
 			{
-				if (autoExpandNode != null)
-				{
-					autoExpandNode.Expand();
-				}
+				autoExpandNode.Expand();
 			}
 
 			autoExpandTimer.Stop();
@@ -61,25 +61,46 @@ namespace Duality.Editor.Controls
 		protected override void OnDragOver(DragEventArgs dragEvent)
 		{
 			base.OnDragOver(dragEvent);
-
 			if (AutoExpandDelay > 0)
 			{
 				Point locationHoverPosition = PointToClient(new Point(dragEvent.X, dragEvent.Y));
 				TreeNodeAdv hoveredNode = GetNodeAt(locationHoverPosition);
+				TreeNodeAdv draggedNode = SelectedNode;
 
-				if ((autoExpandNode == null || autoExpandNode != hoveredNode))
+				// If we hover over ourselves, null or a cihld stop the timer.
+				if (draggedNode == null || hoveredNode == null || draggedNode == hoveredNode || draggedNode.Children.Contains(hoveredNode))
+				{
+					autoExpandTimer.Stop();
+					autoExpandNode = null;
+					return;
+				}
+
+				// If we are different to the currently timing out node, then start it again.
+				if (autoExpandNode != hoveredNode)
 				{
 					autoExpandNode = hoveredNode;
 					StartAutoExpandTimer();
 				}
 			}
+		}
 
+		protected override void OnDragLeave(EventArgs dragLeaveEvent)
+		{
+			base.OnDragLeave(dragLeaveEvent);
+			autoExpandTimer.Stop();
+			autoExpandNode = null;
+		}
+
+		protected override void OnDragDrop(DragEventArgs dragDropEvent)
+		{
+			base.OnDragDrop(dragDropEvent);
+			autoExpandTimer.Stop();
+			autoExpandNode = null;
 		}
 
 		protected override void Dispose(bool disposing)
 		{
 			base.Dispose(disposing);
-
 			autoExpandTimer.Tick -= OnAutoExpandTimerTick;
 		}
 	}
--- a/Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
+++ b/Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
@@ -17,91 +17,80 @@ namespace Duality.Editor.Controls
 
 		public int AutoExpandDelay
 		{
-			get
-			{
-				return autoExpandTimer.Interval;
-			}
-
-			set
-			{
-				autoExpandTimer.Interval = value;
-			}
+			get { return this.autoExpandTimer.Interval; }
+			set { this.autoExpandTimer.Interval = value; }
 		}
 
 		public AutoExpandTreeView(int autoExpandDelay = 1000)
 		{
-			autoExpandTimer.Interval = autoExpandDelay;
-			autoExpandTimer.Tick += OnAutoExpandTimerTick;
+			this.autoExpandTimer.Interval = autoExpandDelay;
+			this.autoExpandTimer.Tick += this.OnAutoExpandTimerTick;
 		}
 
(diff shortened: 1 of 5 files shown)


Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
```diff
@@ -0,0 +1,68 @@
+using Aga.Controls.Tree;
+using System;
+using System.Drawing;
+using System.Windows.Forms;
+
+namespace Duality.Editor.Controls
+{
+	public class AutoExpandTreeView : TreeViewAdv
+	{
+		public int DragOverExpandDelay
```
ilexp: We generally don't do auto-properties in Duality source. Please replace with a regular field-property combo.

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
```diff
@@ -0,0 +1,68 @@
...
+		{
+			get;
+			set;
+		} = 1000;
+
+		private TreeNodeAdv autoExpandNode = null;
```
ilexp: Code Style: All fields as a group, no empty lines between them except to signify larger groups of different categories of fields (none in this case). Also, fields are the first thing in a class definition, followed by properties, and then methods.

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
```diff
@@ -0,0 +1,68 @@
...
+
+		private Timer timer = new Timer();
```
ilexp: "timer" is very general! This is specifically an auto-expand timer. Suggesting to rename this field. Same goes for the Start / Stop methods.

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
```diff
@@ -0,0 +1,68 @@
...
+
+		private void StartTimer()
```
ilexp: See above

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
```diff
@@ -0,0 +1,68 @@
...
+			if (DragOverExpandDelay > 0)
+			{
+				timer.Stop();
+				timer.Interval = DragOverExpandDelay;
+				timer.Tick += OnTimerTick;
+				timer.Start();
+			}
+		}
+
+		private void StopTimer()
```
ilexp: See above

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
```diff
@@ -0,0 +1,68 @@
...
+
+		private Timer timer = new Timer();
+
+		private void StartTimer()
+		{
+			if (DragOverExpandDelay > 0)
+			{
+				timer.Stop();
+				timer.Interval = DragOverExpandDelay;
+				timer.Tick += OnTimerTick;
```
ilexp: You never seem to unsubscribe the event handler again - could be a bug. Please verify. Suggesting to unsubscribe in the `StopTimer` method.

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
```diff
@@ -0,0 +1,68 @@
...
+			set;
+		} = 1000;
+
+		private TreeNodeAdv autoExpandNode = null;
+
+		private Timer timer = new Timer();
+
+		private void StartTimer()
+		{
+			if (DragOverExpandDelay > 0)
```
ilexp: Expanding on the above no-auto-properties comment: When there is no specific reason to do otherwise, please use the backing field directly inside your class. No need to invoke a getter method. This is for all occurrences of get-access of `DragOverExpandDelay`.

ilexp: Something else: All member access (fields, properties, methods) needs to be preceded with a `this.` - this is a code convention in the Duality repository.

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
```diff
@@ -0,0 +1,68 @@
...
+			StopTimer();
+		}
+
+		protected override void OnDragOver(DragEventArgs drgevent)
+		{
+			base.OnDragOver(drgevent);
+
+			if (DragOverExpandDelay > 0)
+			{
+
```
ilexp: No reason to have an empty line here. Please remove for clarity / consistency / readability.

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
```diff
@@ -0,0 +1,68 @@
...
+			timer.Stop();
+		}
+
+		private void OnTimerTick(object sender, EventArgs e)
+		{
+			StopTimer();
+		}
+
+		protected override void OnDragOver(DragEventArgs drgevent)
```
ilexp: There's no reason to abbreviate `drgevent` that strongly (although that just might have been the default name). Suggesting to use `dragEvent` instead.

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
```diff
@@ -0,0 +1,68 @@
...
+		{
+			base.OnDragOver(drgevent);
+
+			if (DragOverExpandDelay > 0)
+			{
+
+				Point p = PointToClient(new Point(drgevent.X, drgevent.Y));
+				var node = GetNodeAt(p);
```
ilexp: Another coding convention: Don't use `var` unless the alternative would decrease readability, such as `foreach`ing over a dictionary with complicated type parameters. This is not one of those cases, please no `var` here :) Also (super high level), suggesting to rename this to `hoveredNode`.

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		protected override void OnDragOver(DragEventArgs drgevent)
+		{
+			base.OnDragOver(drgevent);
+
+			if (DragOverExpandDelay > 0)
+			{
+
+				Point p = PointToClient(new Point(drgevent.X, drgevent.Y));
```
ilexp: Suggesting rename to `localHoverPos` or a similar descriptive name of your choice.

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
ilexp: On second thought, there is no reason to subscribe / unsubscribe that often at all. You could just subscribe the event in the constructor and unsubscribe in a `Dispose` override and just start / stop the timer as needed. A positive side effect would be to produce less garbage due to implicit delegate instantiations.

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
nhold: Looking at this a bit more, it's really just a property exposing the the Interval property of the timer. Do code guidelines prefer just doing the following:

```
private Timer       autoExpandDelayTimer = new Timer();
public int AutoExpandDelay
{
    get
    {
        return autoExpandDelayTimer.Interval;
    }

    set
    {
        autoExpandDelayTimer.Interval = value;
    }
}
```

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
nhold: Yep, we were re-listening every time, possibly degrading performance over time. Now just doing it in a constructor and unsubscribing in `Dispose`. Also added an optional parameter for the constructor to take in the ms for the `autoExpandDelayTimer` interval (1000 is default).

nhold:
> Something else: All member access (fields, properties, methods) needs to be preceded with a this. - this is a code convention in the Duality repository.

Does this include inherited members? I.e `this.PointToClient(...)`?

Other than the above I believe all the code review changes have gone in, plus the following bug fixes\changes:

1. You can no longer open children of the node you are dragging or their children (You can't move a parent to it's children) - If this is expected behaviour let me know.
2. Dragging a node onto a valid 'auto expand' node and then off of the local client area no longer continues to expand the node you last dragged over.
3. If you have already dropped a node the previous node you dragged over will no longer expand afterwards.
4. Fixed crash when moving from project->scene view

There is one bug left to go when dragging project->scene view and vice versa. The behaviour is inconsistent on whether it will expand or not. Not sure if this is something I introduced but when drag and dropping a prefab from project to scene it is created with an empty name.

Source/Editor/DualityEditor/Controls/AutoExpandTreeView.cs
ilexp: Looks good semantically! Style-wise, for property implementations that short, the single-line form is preferred:

```CSharp
private Timer autoExpandDelayTimer = new Timer();

public int AutoExpandDelay
{
    get { return this.autoExpandDelayTimer.Interval; }
    set { this.autoExpandDelayTimer.Interval = value; }
}
```

In other cases, we just omit the newline between getter and setter.

ilexp:
> > Something else: All member access (fields, properties, methods) needs to be preceded with a this. - this is a code convention in the Duality repository.
> 
> Does this include inherited members? I.e `this.PointToClient(...)`?

Yep! All of them.

> Other than the above I believe all the code review changes have gone in, plus the following bug fixes\changes:
> 
> 1. You can no longer open children of the node you are dragging or their children (You can't move a parent to it's children) - If this is expected behaviour let me know.
> 2. Dragging a node onto a valid 'auto expand' node and then off of the local client area no longer continues to expand the node you last dragged over.
> 3. If you have already dropped a node the previous node you dragged over will no longer expand afterwards.
> 4. Fixed crash when moving from project->scene view

Nice work 👍 

> There is one bug left to go when dragging project->scene view and vice versa. The behaviour is inconsistent on whether it will expand or not. Not sure if this is something I introduced but when drag and dropping a prefab from project to scene it is created with an empty name.

Hmm.. I wouldn't know how your changes should introduce that kind of Prefab dragdrop bug, but that's definitely not normal behavior - and I can't reproduce it in an unchanged build. This is how it should look like: https://gfycat.com/KaleidoscopicQuarrelsomeEarthworm

Can you verify whether this is a regression introduced by this PR? _Maybe_ you are somehow accidentally interfering with something that may or may not trigger renaming-related code, leading to emptying the name of a newly instantiated node, but.. just wildly guessing here.

One more issue regarding code structure: Don't do empty lines between methods, except if the method above and the one below belong to different "groups" or "categories of methods. Which grouping you use is up for debate and your judgement, but there should be no empty line without reason. (Collapsed) example:

```CSharp
public AutoExpandTreeView(int autoExpandDelay = 1000)

private void StartAutoExpandTimer()
private void StopAutoExpandTimer()

private void OnAutoExpandTimerTick(object sender, EventArgs e)
protected override void OnDragOver(DragEventArgs dragEvent)
protected override void OnDragLeave(EventArgs dragLeaveEvent)
protected override void OnDragDrop(DragEventArgs dragDropEvent)

protected override void Dispose(bool disposing)
```

I'll let you address the last few issues, but I think we're making good progress here.

> P.s Let me know if I should squash all these commits.

No need to, I'll use the GitHub Squash-Merge feature on this PR.

nhold:
Okay that bug seems to be on master as well (on commit e94e8ca) if you create an empty prefab and drag it into the scene it creates a GameObject with an empty string as a name. Not sure if it's expected behaviour, here it is:

test3

Latest commit employs recursion to go through parents of the hoveredNode to ensure the we aren't dragging a parent over it. I added an extension method for that functionality.

ilexp:
> Okay that bug seems to be on master as well (on commit e94e8ca) if you create an empty prefab and drag it into the scene it creates a GameObject with an empty string as a name. Not sure if it's expected behaviour, here it is:

Ah - good find. I didn't test an empty one. Prefabs remain empty until an existing object is injected into them, but usually they're just created by dragging an object into the Project View. The behavior in your use case is expected, but admittedly not intuitive. Will have to figure out how to deal with this case better.

> On top of that I believe all the stylistic changes are addressed.

They are. Thanks for your contribution and especially for being so positive about change requests! I'll merge this and you will find this update in the next binary release.

PR #520, opened 2017-04-11
Fixed NullReferenceException in MeasureText

ChristianGreiner:
Fixed a NullReferenceException when the parameter of MeasureText (Font.cs) is null


ilexp:
Closed and awaiting cleaned up commit (without the 10 MB sqlite file 😄 ) as discussed on the chat.

PR #525, opened 2017-04-22
Fix GetLocalPosAtTile

mfep:
The current solution works correctly only with the `TilemapRenderer.Origin` set to `TopLeft`. This code works for every alignment case. (I assume that the function should return the tile position in the `TilemapRenderer.GameObj`'s local coordinate system.)


ilexp:
Nice find. Thanks!

Issue #503, opened 2017-02-18
Add Zoom Level Option to Tileset View


mfep:
### Summary

Editing AutoTiles in the Tileset editor is difficult at tiny tile sizes. Although tiles under 24x24 are scaled up, on some displays, editing larger tiles than that is a challenge too.

### Workaround

Windows Magnifier

### Analysis

- The display size of the tiles in the tile editor could be modified arbitrarily
- ..with a pair of buttons on the toolbar,
- and by scrolling the mouse wheel (that's a standard way of resizing the view in many applications)

### Icons
`magnifier_zoom_in` and `magnifier_zoom_out` icons from the fatcow set could be used for the buttons.

ilexp:
Done by @mfep, release in progress.

PR #539, opened 2017-06-03
Manual zoom level support for Tileset View

mfep:
Implements issue #503. The former auto-resize logic has been removed as the zoom level can be set manually.


ilexp:
Thx for the PR 👍 

I was about to start a review on this but then decided to tackle the changes myself, as they were mostly just minor code style ones. Then I got carried away a bit and ended up doing a bit more than I thought. As I still wanted to review your code, I have wrapped my changes into multiple commits, so you can click on each commit and read the changes as an individual review comment. 

Note that the re-introduction of the default zoom level is more than just a nice2have feature, as removing it would have been a "breaking usability" change for the Tile Palette in the Tilemap Editor: Since this one does not have any manual zoom options and relies solely on the default level, it would likely have had serious problems for more extreme tile sizes.

Also, please don't read the number of changes as a critique of your original PR in this case. It was actually quite good, and a lot of these changes only occurred to me in the stream of changes that I was already doing 🙂 

Will merge this soon.

Issue #542, opened 2017-06-18
Backwards tab order in New Project dialog.


misnor:
### Summary

Tab order appears to be backwards on the new project dialog.

### How to reproduce [Bugs]

1. Open Duality Editor
2. Click "Project Folder" text box
3. Hit Tab, repeat hitting tab.

### Workaround [Bugs]

1. Open Duality Editor
2. Click "Project Folder" text box
3. Hit **SHIFT** + Tab, repeat hitting **SHIFT** + Tab.

### Attachments

[Gif illustrating tab order in New Project](http://i.imgur.com/3BLFceE.gif)

Relatively minor issue, something i noticed right away when trying out Duality.

Windows 10 Pro 64-bit (10.0, Build 15063) if needed.

ilexp:
Fixed in PR #543.

PR #543, opened 2017-06-18
Fix for backwards tab order on new project dialog (#542)

misnor:
Potential fix for minor issue #542, All tests are still passing after changing the tab order of the form. (Within the code its self the tab order does look like nonsense, however, the tab order on the form is now correct and behaves as expected.

All tests remain passing.


Source/Editor/DualityEditor/Forms/NewProjectDialog.Designer.cs
```diff
@@ -63,6 +63,7 @@ private void InitializeComponent()
 			// 
 			this.folderView.BackColor = System.Drawing.Color.FromArgb(((int)(((byte)(196)))), ((int)(((byte)(196)))), ((int)(((byte)(196)))));
 			this.folderView.BorderStyle = System.Windows.Forms.BorderStyle.FixedSingle;
+			this.folderView.ColumnHeaderHeight = 0;
```
ilexp: This line wouldn't have been necessary for this commit - it's not crucial, but always best to keep out unintended changes :)

ilexp:
Hey and thanks for the PR! Merged.

PR #560, opened 2017-08-22
Fixed Time.GameTimer related bug in DualityApp.Update

mfep:
*Time has to pass.*


ilexp:
Release in progress. Should be out there in about 15 minutes.

Issue #538, opened 2017-06-01
Fix IUniqueIdentifyable Implementations to Not Rely on GetHashCode


ilexp:
### Summary

Right now, Component and GameObject rely on GetHashCode to produce the same result for the same value across different runs. While true for now, this is not guaranteed to be true across runtimes and could one day cause problems. Change the implementations to not rely on GetHashCode.

### Analysis

- It's a problem specifically in the case of `IUniqueIdentifyable`, because the returned value of `PreferredId` is used for persistence.
- It will not cause errors, but may cause a complete restructuring of Scene and Prefab files on every save, killing version control usability.
- Spotted in the following code:
  - [GameObject](https://github.com/AdamsLair/duality/blob/bb9ad80b32a62c5ad9d2ed977ba66f8b72f8a0e6/Source/Core/Duality/GameObject.cs#L253-L256)
  - [Component](https://github.com/AdamsLair/duality/blob/bb9ad80b32a62c5ad9d2ed977ba66f8b72f8a0e6/Source/Core/Duality/Component.cs#L136-L150)
  - Others? Investigate!

SirePi:
Grabbed!

ilexp:
> Grabbed!

👍 

Can't assign you for some reason, but consider yourself assigned.

SirePi:
Implementation based roughly on this https://stackoverflow.com/questions/2351087/, tested with 1M calculation runs show an increase of roughly 50% in execution time comparing the new algorithm vs the standard GetHashCode implementation for strings.

For GUIDs, execution time is now comparable to the old GetHashCode for strings.

ToDo: unit tests to be implemented before delivery

ilexp:
Done by @SirePi.

PR #572, opened 2017-10-04
New unique identifier

SirePi:
Implementing issue #538

### DIFF
--- a/Source/Core/Duality/Backend/DefaultPluginLoader.cs
+++ b/Source/Core/Duality/Backend/DefaultPluginLoader.cs
@@ -5,6 +5,7 @@ using System.Reflection;
 using System.Collections.Generic;
 
 using Duality.IO;
+using System.Text;
 
 namespace Duality.Backend
 {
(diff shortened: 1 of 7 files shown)


Source/Core/Duality/Backend/DefaultPluginLoader.cs
```diff
@@ -99,42 +100,53 @@ public void Init()
...
-					Environment.CurrentDirectory,
-					Environment.CommandLine,
-					osName + (osFriendlyName != null ? (" (" + osFriendlyName + ")") : ""),
-					Environment.Is64BitOperatingSystem,
-					Environment.Is64BitProcess,
-					Environment.Version,
-					Environment.ProcessorCount);
+
+				StringBuilder envInfo = new StringBuilder();
+				envInfo.AppendLine("Environment Info:")
```
ilexp: - 👍 for refactoring to use `StringBuilder` in this case.
- Please don't use fluent syntax unless there's a very good reason to.
- Consider (not sure, you decide) using `AppendFormat(...)` + `AppendLine()` combo instead of repeated `string.Format`, might otherwise allocate as much as previous version without `StringBuilder`.

Source/Core/Primitives/Math/MathF.cs
```diff
@@ -1246,6 +1246,34 @@ public static void Swap<T>(ref T first, ref T second)
 		/// <param name="baseHash"></param>
 		/// <param name="otherHash"></param>
 		/// <returns></returns>
+		public static void CombineHashCodeUint(ref uint baseHash, uint otherHash)
```
ilexp: - I think you don't need `uint` versions of `CombineHashCode` if you just use the `int` versions and do a cast to `uint` inside an `unchecked` block.
- Please remove `uint` overloads to keep maintenance cost / gain ratio lower.

Test/Core/Serialization/SerializerTest.cs
```diff
@@ -566,6 +566,26 @@ public SerializerTest(Type format)
 			}
 		}
 
+		[Test] public void CalcUniqueIdentifier()
```
ilexp: 👍 for testing.

ilexp:
Just noticed your other PRs changes are in this one too. Sorry for the confusion on the review side. Next time, please keep them in separate branches though 😄

SirePi:
Weird.. I have 2 branches though :/
Ok for the changes.

ilexp:
> EDIT: are you sure about the uint/int situation though? 

85% sure, maybe double-check the unchecked conversion in debug mode to verify.

> What's the point of having `IUniqueIdentifyable.PreferredId` as uint if so?

Mostly because negative object IDs look weird in an XML or like an error in binary.

SirePi:
Can safely ignore `MathF.cs`, it was basically reverted back to the original one

Issue #533, opened 2017-05-12
Write RigidBody / Physics Unit Tests


ilexp:
### Summary

There are already unit tests that check some basic cloning / copying functionality, but there are none that test actual physics and transformation behavior. Add a few basic tests to ensure physics works properly.

### Analysis

- The basic test layout could be to create a scene, make it current and simulate a certain number of frames with a fixed timestep, similar to the sandbox single-frame step.
- Tests could include:
  - A ball falling down onto a platform.
  - Non-static bodies not depending on parent transform, unless `AllowParent` was set to true.
  - Probably lots of others, including properties, joints, etc., but keep it short and simple for now.

Barsonax:
I could pick this one up. Already did some unit tests in Pathfindax that used raycasts. I can add the tests to RigidBodyTest right?

Not really sure what you mean with 'Non-static bodies not depending on parent transform, unless AllowParent was set to true.' though

ilexp:
> I could pick this one up. Already did some unit tests in Pathfindax that used raycasts. I can add the tests to RigidBodyTest right?

Yes, that would be a good place. As the number of tests grows, we might think about splitting it up, but not now.

> Not really sure what you mean with 'Non-static bodies not depending on parent transform, unless AllowParent was set to true.' though

If you parent a GameObject that has a RigidBody to another one with a Transform, the child object will ignore changes to its parent Transform to ensure a well-defined physics simulation. This exception will be disabled when switching on `AllowParent`.

Maybe just start with the "A ball falling down onto a platform" one for now, do a PR for that one only, we'll do the reviews and then go from there.

ilexp:
There are existing tests that call `DualityApp.Update`, which do seem to work fine on my machine and the CI server, for example `SceneTest.SwitchToDuringLeave` and others in the same class. Can you step through them and see if you get the same behavior?

Barsonax:
Tested with `SceneTest.SwitchToDuringLeave` and this still happens:
<img width="981" alt="image" src="https://user-images.githubusercontent.com/19387223/31314853-96ec84c0-ac0b-11e7-8b03-0dd8d7a77da4.png">

ilexp:
I'll try to see if I can reproduce this with any of the VS versions I have installed. I'm usually on Visual Studio 2013, maybe this only happens for newer versions.

Barsonax:
Iam using Visual studio 2017 and running the tests using the Resharper test runner.

ilexp:
[...]

PR #575, opened 2017-10-08
#533 added FallingBallOnPlatform test

Barsonax:


Test/Core/Components/CollisionListener.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+
+namespace Duality.Tests.Components
+{
+	/// <summary>
+	/// A collision listener to expose the collision events to tests
+	/// </summary>
+	public class CollisionListener : Component, ICmpCollisionListener
```
ilexp: Rename to `CollisionEventReceiver` in order to match `InitializableEventReceiver` [here](https://github.com/AdamsLair/duality/blob/f72221bc75e27a2450d96915adb3513c5df94284/Test/Core/Components/InitializableEventReceiver.cs), which has a similar role in testing scenarios.

Test/Core/Components/CollisionListener.cs
```diff
@@ -0,0 +1,29 @@
...
+	{
+		public event EventHandler<CollisionEventArgs> CollisionBegin;
```
ilexp: In the same vein, prefer internally storing received events in a list for later validation over redirecting events. Probably need a custom local struct for this, containing event type (local enum) and args.

Test/Core/Components/RigidBodyTest.cs
```diff
@@ -1,21 +1,54 @@
-﻿using System;
```
ilexp: While removing unused usings is good, please keep using directives grouped and the groups separated with a single newline. `NUnit.Framework` is a different group than the `Duality` usings, which is a different one than the `System` usings.

Test/Core/Components/RigidBodyTest.cs
```diff
@@ -1,21 +1,54 @@
...
 using NUnit.Framework;
 
 namespace Duality.Tests.Components
 {
 	[TestFixture]
 	public class RigidBodyTest
 	{
+		[Test] public void FallingBallOnPlatform()
+		{
+			var scene = new Scene();
```
ilexp: Don't use `var` if it can be reasonably avoided, keep variable definitions explicit instead.

Test/Core/Components/RigidBodyTest.cs
```diff
@@ -1,21 +1,54 @@
...
+			Scene.SwitchTo(scene);
+
+			//Create the ball
```
ilexp: Missing space after the `//`

Test/Core/Components/RigidBodyTest.cs
```diff
@@ -1,21 +1,54 @@
...
+			var ball = new GameObject("Ball");
+			var ballTransform = ball.AddComponent<Transform>();
+			var ballBody = ball.AddComponent<RigidBody>();
+			var listener = ball.AddComponent<CollisionListener>();
+			bool onCollisionBeginCalled = false, onCollisionEndCalled = false, onCollisionSolveCalled = false;
```
ilexp: See comment on collision listener test class, replace flags set in inline event handlers with list storage.

Test/Core/Components/RigidBodyTest.cs
```diff
@@ -1,21 +1,54 @@
...
+			var platformTrans = platform.AddComponent<Transform>();
+			var platformBody = platform.AddComponent<RigidBody>();
+			platformBody.AddShape(new ChainShapeInfo(new[] { new Vector2(-1, 0), new Vector2(1, 0), }));
+			platformBody.BodyType = BodyType.Static;
+			scene.AddObject(platform);
+			platformTrans.Pos = new Vector3(0, 5, 0);
+
+			for (int i = 0; i < 6; i++)
+			{
+				Thread.Sleep(16); //Simulate ~60 fps.
```
ilexp: Don't sleep! Instead, we can modify the Duality core to expose the same fixed-step time simulation that we get in the editor. Take a look at `DualityApp` around [here](https://github.com/AdamsLair/duality/blob/f72221bc75e27a2450d96915adb3513c5df94284/Source/Core/Duality/DualityApp.cs#L550). The `EditorUpdate` method has a `forceFixedStep` parameter. Add it to `Update` as well and use it in the test.

To make this not a breaking change, introduce a parameterless `Update` overload to `DualityApp` that simply calls the modified one with the new parameter being false.

Test/Core/Components/RigidBodyTest.cs
```diff
@@ -1,21 +1,54 @@
...
+				DualityApp.Update();
+			}
+			Assert.IsTrue(ballTransform.Pos.Y > 0);
+			Assert.IsTrue(ballTransform.Pos.X == 0);
```
ilexp: Allow a bit of flexibility here. If, for some reason, a future change in physics would make `X` be -0.00263 or so, that would still be totally fine. Come up with a reasonably small threshold and use it!

Source/Core/Duality/DualityApp.cs
```diff
@@ -526,12 +526,13 @@ public static void ApplyResolution(int width, int height, bool fullscreen)
 		/// <summary>
 		/// Performs a single update cycle.
 		/// </summary>
-		public static void Update()
+		/// <param name="forceFixedStep">If true use a timestep thats equal to <see cref="Time.MsPFMult"/> for the update</param>
+		public static void Update(bool forceFixedStep = false)
```
ilexp: We actually need two different Update overloads for compatibility reasons. While the default parameter allows for source compatibility, it still breaks binary compatibility: A package depending on this method signature would need to be recompiled in order to work again. Using a default parameter value is something that can be done in v3.0, but not here.

Test/Core/Components/CollisionEventReceiver.cs
```diff
@@ -0,0 +1,41 @@
...
+			public readonly CollisionEventArgs Args;
+
+			public CollisionEvent(Component sender, CollisionEventArgs args)
+			{
+				this.Sender = sender;
+				this.Args = args;
+			}
+		}
+
+		public readonly List<CollisionEvent> CollisionBegin = new List<CollisionEvent>();
```
ilexp: With three separate lists instead of a struct with a local "event type" enum, the order in which different events arrive is lost. Not a blocker for a merge (read as FYI, you decide), but probably not future-proof either.

Test/Core/Components/RigidBodyTest.cs
```diff
@@ -16,6 +13,46 @@ namespace Duality.Tests.Components
...
+
+			// Check if the position is within expected values
+			Assert.IsTrue(ballTransform.Pos.Y > 3f);
+			Assert.IsTrue(ballTransform.Pos.Y < 5f);
+			Assert.IsTrue(Math.Abs(ballTransform.Pos.X) < 0.01f);
+			Assert.IsTrue(Math.Abs(ballTransform.Pos.Z) < 0.01f);
+
+			// Check if the collision events were triggered
+			Assert.IsTrue(listener.CollisionBegin.Count == 1);
+			Assert.IsTrue(listener.CollisionEnd.Count == 0); //The ball should keep touching the platform once it collides
```
ilexp: 👍 for the explaining comment, but there's that missing space again 😄

Test/Core/Components/RigidBodyTest.cs
```diff
@@ -16,6 +13,46 @@ namespace Duality.Tests.Components
...
+			Assert.IsTrue(listener.CollisionSolve.Count == 1);
```
ilexp: Does this remain true when increasing the number of simulation frames in the loop above? Solve is called once per frame as long as collisions are resolved / bodies are not sleeping. Seems a bit flaky to require it to be really equal to one. Maybe just check for `>= 1`.

Source/Core/Duality/DualityApp.cs
Barsonax: Didn't realized this would break binary compatibility. Ill change it at once.

Test/Core/Components/CollisionEventReceiver.cs
Barsonax: Ye the order of the events could be something that's important, good point. Actually when I think about it might as well check that in the ball test too? Pretty sure begin collision has to come before the others.

Test/Core/Components/RigidBodyTest.cs
Barsonax: Didn't realized Solve is called per frame so Ill change that.

Test/Core/Components/RigidBodyTest.cs
Barsonax: Note to self: write space inserting comment AI

Test/Core/Components/CollisionEventReceiver.cs
ilexp: > Actually when I think about it might as well check that in the ball test too? Pretty sure begin collision has to come before the others.

Yep, makes sense. If you check for this as well, the intended order would be `Begin`, `Solve` times N (any N is fine), `End`, and that order might show up multiple times in case the ball bounces off. For the simple test, maybe just set the `Restitution` values of both ball and platform to zero, so there should be no bouncing.

Barsonax: Restitution is 0 by default but I will set it explicitly nonetheless in case that changes.

Barsonax: Added a DebuggerDisplay attribute to CollisionEvent as well so you can see at once the type and the gameobject it collided with. If you prefer a ToString override to achieve this I can change that.

Issue #291, opened 2016-02-18
Fix the Rect Version of RigidBody.PickShapesGlobal


ilexp:
The current implementation of [RigidBody.PickShapes](https://github.com/AdamsLair/duality/blob/master/Duality/Components/Physics/RigidBody.cs#L717-L768) is hacky at best and [doesn't work properly](http://forum.adamslair.net/viewtopic.php?p=4272#p4272). 

- Write a unit test that proves the implementation doesn't work in all cases.
- Implement a version that works, and (ideally) does so faster.

adambiltcliffe:
Test project demonstrating errors in the implementation: [PickShapesBugDemo.zip](https://github.com/AdamsLair/duality/files/139614/PickShapesBugDemo.zip)

Barsonax:
I will take a look at this, since the last comment is over a year ago I guess this one is free to pick up.

ilexp:
It is! Good luck 👍

Barsonax:
I have reproduced a situation in a unit test. I do not understand RigidBody.PickShapes but seeing as you call the current implementation hacky I guess that kind of makes sense.

From what I understand is that PickShapesGlobal first finds rigidbodies that could potentially be in the rectangle but then it still has to weed out the false positives which it does with PickShapes. All PickShapes has to do is check each shape in the rigidbody and see if it is in the rectangle or not. So first you would have to check what kind of shape it is and then do the check for that type.

ilexp:
> From what I understand is that PickShapesGlobal first finds rigidbodies that could potentially be in the rectangle but then it still has to weed out the false positives which it does with PickShapes. All PickShapes has to do is check each shape in the rigidbody and see if it is in the rectangle or not. 

Correct.

> So first you would have to check what kind of shape it is and then do the check for that type.
>
> This could be achieved by adding a abstract contains method to ShapeInfo which the shapes have to implement. A alternative could be with casting but I dont like that.

That would be a solution, but [Farseer](https://github.com/AdamsLair/farseerduality) already should have logic for this and I'd prefer to use that instead of rolling your own. Maybe investigate where that logic exists and see how it can be used in that case first. Writing it from scratch shouldn't be the first approach.

Barsonax:
Ok I will check that out.

I will keep the abstract method on ShapeInfo though and put the farseer code in there.

Barsonax:
[...]

PR #577, opened 2017-10-12
#291 Fixing pickshapes code

Barsonax:
#291 

Tested the code using the unit tests I made and opened up a duality instance and used code in the https://github.com/AdamsLair/duality/files/139614/PickShapesBugDemo.zip to test this. Could not find cases where I got false results except when changing the scale of the transform while the game is running but this is another issue and this PR is quite big already.


ilexp:
No review yet, but code-related continued discussion from issue #291:

> Also compared the performance and the new version is actually slower currently. It varies but I have seen cases where its twice as slow. Ill see if I can get it faster.
>
> EDIT: after looking through the old implementation it seems to be doing only collisions between the AABB's? No wonder its faster.

Can you profile to see where the bottleneck is and optimize the new version?

Barsonax:
Currently going through the code to fix any unnecessary calculations and do a bit of cleanup.

Barsonax:
It seems a lot of performance is being eaten by GetWorldPoint and GetLocalPoint. Upon closer inspection it calls into TransformCoord which does a Math.Sin and a Math.Cos which are quite expensive operations (over 100x slower than multiplication in my own testing).

On a side note isn't it possible to just multiply the vector with a transformation matrix? I expect that to be a lot faster.

ilexp:
Wait a sec - you didn't seem to use the [Farseer methods](https://github.com/AdamsLair/farseerduality/blob/65b1396bf03f83706fe740c097e56b25053d058a/Collision/Collision.cs#L761) for the check [as we discussed earlier](https://github.com/AdamsLair/duality/issues/291#issuecomment-335547689) - what's the reason for this? I know the contact checks themselves were inaccessible, but the static versions are, right? Pretty sure they're already optimized well and I don't want to duplicate this functionality if we already have it just sitting there.

ilexp:
Took a quick look at Farseer code to see how those methods are used and make sure I'm not asking for something stupid 😄

Seems like `Manifold` and `Transform` are public, and you can get each bodies `Transform` using its public `GetTransform` method. So theoretically, you could [do the same as Farseer does](https://github.com/AdamsLair/farseerduality/blob/65b1396bf03f83706fe740c097e56b25053d058a/Dynamics/Contacts/Contact.cs#L293), except not using `Xf` directly, but doing a detour through `GetTransform`. 

Using this approach, we'd just re-use the existing Farseer code and existing data structures, so there would be no need for any coordinate transformation beyond that needed for a constructing a single temporary `PolygonShape` in the beginning of the method. With that, there's also be no need for a separate `Box` class, and we'd have the code in place to extend the collision check easily in the future - for example towards non-rect shapes.

API-wise, you could operate directly on the Farseer bodies and iterate their fixtures / shapes, so we wouldn't even need any changes to existing `ShapeInfo` API, which would lower the complexity of the PR a bit too.

Barsonax:
Its not clear to me how I could use the farseer code here. I tried it for polygons but it did not detect any collisions.

Barsonax:
I have implemented the PickShapes using farseer for polygons and circles. See #579

ilexp:
Looks much better :) Closing this PR, as it is superseded by #579.

Issue #570, opened 2017-10-03
Consider Incorporating FarseerDuality Project Into Main Package / Solution


ilexp:
### Summary

Duality uses a [custom Farseer Physics](https://github.com/AdamsLair/farseerduality) port, which is built and deployed separately. This not only increases iteration times heavily, but it also makes Duality-specific specialization harder. Consider moving the `FarseerDuality` project into main and renaming it to `Duality.Physics`, paving the way for gradually improving and streamlining it for integrated Duality usage.

### Analysis

- Reasoning behind moving the port closer to Duality and away from its origin:
  - It's based on an old Farseer version (3.3.1) and updating to the head version is unlikely because the gain / stability tradeoff doesn't sound great with regard to [unresolved issues](https://farseerphysics.codeplex.com/workitem/list/basic) in the latest stable (3.5) of the original project which could easily turn out to be critical.
  - Farseer will no longer be updated and Velcro will take a bit until reaching stability, so a switch is unlikely at this point.
  - The custom Duality port of Farseer has already deviated from the original a lot.
  - It has also proven itself to be very stable over the past years.
  - With the above in mind, further deviation and specialization towards Duality might be beneficial.
  - As there will be no direct merge between core and physics library, switching to a different physics library internally will always remain a possibility.
- The project should probably be located in `Source/Core/Physics` next to `Primitives` and `Duality`.
  - Could consider renaming the folders to `Duality.Physics` and `Duality.Primitives`, but should apply the same naming convention to the `Source/Editor` folders as well, making them `Duality.Editor` and `Duality.Updater`.
  - This would prompt renaming their project files as well. Lots of renames, but then project names, package names and namespaces would be aligned.
- Even when moving it to the main repository, physics should remain a separate project.
- Efficiency / performance improvement akin to https://github.com/VelcroPhysics/VelcroPhysics/issues/29 can be considered in future iterations.

Jared-Miller:
Started merging the Farseer code into the main Duality project. For now, I left all the package names as is and put the Farseer code into Source/Core/Physics named DualityPhysics, to match the way primitives was structured. 

@ilexp 

- How would you prefer to handle the assembly information for the merged project?
- Do you want the global .editorconfig styling changes applied to the DualityPhysics project?

ilexp:
> How would you prefer to handle the assembly information for the merged project?

[...]

PR #579, opened 2017-10-13
#291 Fixing Pickshapes using farseer

Barsonax:


ilexp:
Since there's a ToDo comment, is this still WiP? Let me know when it's ready for review 👍

Barsonax:
The chainshape part is buggy so I left that out for now.

ilexp:
I'll take a look as soon as I get around to it. Since polygons and circles only would be equal to the old functionality, we could merge this PR even without the chain shape checks after review. Should be fine as long as we don't introduce a regression.

Source/Core/Duality/Components/Physics/RigidBody.cs
```diff
@@ -772,50 +775,65 @@ public bool PickShapes(Vector2 worldCoord, Vector2 size, List<ShapeInfo> pickedS
...
 				{
-					pickedShapes.Add(s);
-					continue;
+					Collision.CollidePolygons(ref manifold, boxShape, ref boxTransform, polygonShape, ref bodyTransform);
+					if (manifold.PointCount > 0)
+					{
+						pickedShapes.Add(s);
+						continue;
+					}
 				}
```
ilexp: If it's a polygon shape, but does not have any collisions, it will continue checking whether it could be a circle shape next. Consider changing this to an if-else structure, or moving the continue statement out of the point count check!

Source/Core/Duality/Components/Physics/RigidBody.cs
```diff
@@ -772,50 +775,65 @@ public bool PickShapes(Vector2 worldCoord, Vector2 size, List<ShapeInfo> pickedS
 		{
 			if (this.body == null) return false;
 
-			Vector2 fsWorldCoord = PhysicsUnit.LengthToPhysical * worldCoord;
-			FarseerPhysics.Collision.AABB fsWorldAABB = new FarseerPhysics.Collision.AABB(fsWorldCoord, PhysicsUnit.LengthToPhysical * (worldCoord + size));
-
+			PolygonShape boxShape = new PolygonShape(new Vertices(new List<Vector2>() {
```
ilexp: Glad you didn't bother keeping the old `fsWorldCoord` naming / variable scheme. That was terrible. This one looks definitely cleaner.

Source/Core/Duality/Components/Physics/RigidBody.cs
```diff
@@ -772,50 +775,65 @@ public bool PickShapes(Vector2 worldCoord, Vector2 size, List<ShapeInfo> pickedS
...
+				PhysicsUnit.LengthToPhysical * worldCoord,
+				PhysicsUnit.LengthToPhysical * new Vector2(worldCoord.X + size.X, worldCoord.Y),
+				PhysicsUnit.LengthToPhysical * (worldCoord + size),
+				PhysicsUnit.LengthToPhysical * new Vector2(worldCoord.X, worldCoord.Y + size.Y) }), 1);
+			Mat22 rot = new Mat22();
```
ilexp: Please add a newline, so the box transform creation is its own block, similar to the body transform retrieval below. Also, maybe move the oldCount and manifold declarations down to directly in front of the loop, so it doesn't get mixes up with the transform creation blocks?

Source/Core/Duality/Components/Physics/RigidBody.cs
```diff
@@ -772,50 +775,65 @@ public bool PickShapes(Vector2 worldCoord, Vector2 size, List<ShapeInfo> pickedS
...
+			rot.SetIdentity();
+			Vector2 pos = new Vector2(0, 0);
+			FarseerPhysics.Common.Transform boxTransform = new FarseerPhysics.Common.Transform(ref pos, ref rot);
```
ilexp: Are transform position and rotation public fields? If so, no need to declare and assign rotation and position separately! Just call SetIdentity directly on the box transform rotation field, and assign the position directly without declaring an extra variable. Could also use the default / empty constructor in that case.

(If they are not public fields, disregard my comment.)

Source/Core/Duality/Components/Physics/RigidBody.cs
```diff
@@ -772,50 +775,65 @@ public bool PickShapes(Vector2 worldCoord, Vector2 size, List<ShapeInfo> pickedS
...
+				CircleShape circleShape = f.Shape as CircleShape;
+				if (circleShape != null)
 				{
-					if (f.TestPoint(ref fsTemp))
+					Collision.CollidePolygonAndCircle(ref manifold, boxShape, ref boxTransform, circleShape, ref bodyTransform);
+					if (manifold.PointCount > 0)
 					{
 						pickedShapes.Add(s);
-						break;
+						continue;
```
ilexp: See above!

Source/Core/Duality/Components/Physics/RigidBody.cs
```diff
@@ -772,50 +775,65 @@ public bool PickShapes(Vector2 worldCoord, Vector2 size, List<ShapeInfo> pickedS
...
+				// This one is still buggy
+				//var chainShape = f.Shape as ChainShape;
+				//if (chainShape != null)
+				//{
+
+				//	for (int j = 0; j < chainShape.Vertices.Count - 1; j++)
+				//	{
+				//		var jNext = j + 1;
+				//		if (jNext >= chainShape.Vertices.Count) jNext = 0;
+				//		var edgeShape = new EdgeShape(chainShape.Vertices[j], chainShape.Vertices[jNext]);
```
ilexp: Not sure if that will fix it, but instead of creating a new edge shape every loop iteration, create it once and let the chain shape configure it. There's a `GetChildEdge` method that is public and I believe also used by Farseers internals, [here](https://github.com/AdamsLair/farseerduality/blob/65b1396bf03f83706fe740c097e56b25053d058a/Collision/Shapes/ChainShape.cs#L128).

Alternatively, if you want to keep chain shapes our of this PR for now, maybe just add a note to the commented out section regarding this, for future reference.

Source/Core/Duality/Components/Physics/RigidBody.cs
Barsonax: Ye missed that, it should obviously not do the circle check after it already done the polygon one.

Too bad we dont have C# 7 as pattern matching would make this so much more readable.

Source/Core/Duality/Components/Physics/RigidBody.cs
Barsonax: I was still experimenting with this part. I already tried it with GetChildEdge but got some weird results so have to investigate that some more. Not for this PR though so Ill put a clarifying comment there

Barsonax:
When I get back home I will give you the setup I used for measuring performance. So far the performance was slower than my own implementation.

Good point about the private parameterized instance method I will add that.

Source/Core/Duality/Components/Physics/RigidBody.cs
Barsonax: Found out there is a ShapeType enum property on the fixture so I can use that in a switch. Much more readable.

Barsonax:
Tested the performance and compared to my own implementation the farseer is slower. How much depends on the shape and where the collision is:
-Circles are ~2x as slow. The performance is pretty much constant no matter where you touch the circle
-Simple polygons (3 vertices). are 1.2x-3x as slow. The farseer version seems to have a constant performance no matter where you touch the polygon while in my own implementation it matters what edge or vertice you touch. 
-Complex polygons (14 vertices). The farseer version seems to again have a constant performance no matter where you touch the polygon. The performance of my own implementation seems to vary as much as 10x depending on where you touch the polygon. Roughly speaking they have the same performance but farseer is more balanced.

Now the question is is it worth sacrificing some performance to avoid duplicate collision code? Note that my own implementation is not optimized yet. I think it should be possible to operate directly on the farseer data now I understand farseer a little better for instance.

Tbh I don't think anyone will be running in performance issues anytime soon with PickShapes. 

Attached is the Data and Source folder I used when testing this. To use:
1. Copy the Data and Source folder to the duality/Build/Output
2. Build the project in the branch you want to test (this one or the other one with my implementation)
3. Debug the ProjectPlugins solution
4. The time it took to perform the PickShapesGlobal is printed to the output. Note that PickShapesGlobal is called 50000 times per update in order to make the performance more visible.

[PickshapesTest.zip](https://github.com/AdamsLair/duality/files/1384835/PickshapesTest.zip)

Source/Core/Duality/Components/Physics/RigidBody.cs
```diff
@@ -772,56 +777,60 @@ public bool PickShapes(Vector2 worldCoord, Vector2 size, List<ShapeInfo> pickedS
...
+				PhysicsUnit.LengthToPhysical * worldCoord,
+				PhysicsUnit.LengthToPhysical * new Vector2(worldCoord.X + size.X, worldCoord.Y),
+				PhysicsUnit.LengthToPhysical * (worldCoord + size),
+				PhysicsUnit.LengthToPhysical * new Vector2(worldCoord.X, worldCoord.Y + size.Y) 
+			}), 1);
+
+			FarseerPhysics.Common.Transform boxTransform = new FarseerPhysics.Common.Transform();
+			boxTransform.SetIdentity();
+			
+			return PickShapes(boxShape, boxTransform, pickedShapes);
```
ilexp: Please use `this.` for instance member access as well. Would merge anyway since it's really just a detail, so whoever of us gets to it first can do it.

ilexp:
Changes look good so far!

> Now the question is is it worth sacrificing some performance to avoid duplicate collision code?

Good question! In this case, I think the answer should be "yes, if it's not too severe", but it should definitely be on a list of potential optimization for when issue #570 is done.

Will take a closer look at performance myself before merging - if it doesn't have a serious impact except in benchmarks, I think it should be fine. Will have to see how close it is to real-world impact.

ilexp:
> Note that there are currently no unit tests in this branch. I could copy some over from #577.

Took a brief look and probably wouldn't do a merge for them as-is, so I'd keep this PR separate for now so we can get it merged. Tests for this would be great, but I'd probably do a draft / design pass in the discussion of #533 first before implementing it.

----

Found a small bug where a ShapeInfo was sometimes added multiple times to the result list for complex polygons and fixed it. 

Looked into performance and getting between 30ms and 350ms for 50000 tests / up to 0.007 ms per test on my machine depending on which bodies are checked - not great, but not catastrophic either. I do see some optimization potential on the Farseer side of things, but that's an issue for another day.

A bit busy right now, so will have to do it later, but I'd say this is a merge. Nice work 👍 

### ToDo:
- After the merge, add a note to #291 stating that there's still the chain version to do, and link to the code segment in question.
- Add a note to #533 for potential to add unit tests for this as well and briefly discuss the where / which / how.

Barsonax:
Hmm well the contains check also adds overhead in all cases atm. Not sure what would be faster since that will also depend on how much shapes you are picking. This would need benchmarking.

Changing the signature would break compatibility so definitely not for v2.

ilexp:
> Hmm well the contains check also adds overhead in all cases atm. Not sure what would be faster since that will also depend on how much shapes you are picking. This would need benchmarking.

Yep, but also different kinds of overhead! Perf right now, but design in the other case, e.g. forcing the callsite to use a more complex collection. Totally agree on the perf consideration though, if it shows up on the profiler at some point, this will have to change.

Issue #337, opened 2016-06-24
Provide Proper Error Messages When Using Multi-Dimensional Arrays


ChristianGreiner:
#### Summary

Duality throws an InvalidCastException, when you create a byte array in a component:

`private byte[,] tiles = new byte[16, 16];
`
#### Attachments

Exception:

```
[Edit] Error:   An error occurred: InvalidCastException: Unable to cast object of type 'System.Byte[]' to type 'System.Byte[,]'.
CallStack:
   at Duality.Cloning.CloneProvider.Duality.Cloning.ICloneOperation.HandleObject[T](T source, T& target) in c:\projects\duality\Duality\Cloning\CloneProvider.cs:line 837
   at Duality.Cloning.CloneProvider.PerformCopyChildObject(Object source, Object target, CloneType typeData) in c:\projects\duality\Duality\Cloning\CloneProvider.cs:line 602
   at Duality.Cloning.CloneProvider.Duality.Cloning.ICloneOperation.HandleObject[T](T source, T& target) in c:\projects\duality\Duality\Cloning\CloneProvider.cs:line 814
   at Duality.Cloning.ExtMethodsICloneOperation.HandleObject[T](ICloneOperation operation, T source, T target) in c:\projects\duality\Duality\Cloning\ICloneOperation.cs:line 22
   at Duality.Cloning.CloneProvider.PerformCopyObject(Object source, Object target, CloneType typeData) in c:\projects\duality\Duality\Cloning\CloneProvider.cs:line 518
   at Duality.Cloning.CloneProvider.CloneObject[T](T source, Boolean preserveCache) in c:\projects\duality\Duality\Cloning\CloneProvider.cs:line 130
   at Duality.Editor.UndoRedoActions.CreateComponentAction.<Do>b__0(Component o) in c:\projects\duality\DualityEditor\UndoRedoActions\CreateComponentAction.cs:line 47
   at System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Duality.Editor.UndoRedoActions.CreateComponentAction.Do() in c:\projects\duality\DualityEditor\UndoRedoActions\CreateComponentAction.cs:line 47
[...]

PR #580, opened 2017-10-15
#337

Barsonax:


Barsonax:
Oh it did those changes when I build the project. @ilexp is this some generated file if so should it even be in source control?

Barsonax:
Looked in the other sample projects and the same *.core.xml file was removed from source control there so I did the same for BasicMenu.core.xml

ilexp:
> Looks good.. don't understand what the changes to BasicMenu have to do with this though..

👍 

> Looked in the other sample projects and the same *.core.xml file was removed from source control there so I did the same for BasicMenu.core.xml

👍

Issue #581, opened 2017-10-15
Packagemanager tests does not handle spaces in the path


Barsonax:
### Summary

If there is a space in the path to your output folder the unit tests for the package manager will fail. 
This is because PackageManager.CreateRepository is replacing the spaces with a %20 which in turn makes the repository unable to find the path. The Uri class is responsible for this as the Uri.AbsolutePath is putting these %20 in the repositoryUrl. See [PackageManager.CreateRepository](https://github.com/AdamsLair/duality/blob/master/Source/Editor/DualityEditor/PackageManagement/PackageManager.cs#L689)

The solution is to unescape the uri with Uri.UnescapeDataString.

### How to reproduce [Bugs]

- Clone duality and make sure the path to the output folder has a space somewhere in its name
- Try to run the package manager unit tests (such as GetPackage)
- Profit

### Workaround [Bugs]

- Make sure you don't have a space in the path to the output folder

ilexp:
Fixed by in PR #581.

PR #582, opened 2017-10-15
#581

Barsonax:


Source/Editor/DualityEditor/PackageManagement/PackageManager.cs
```diff
@@ -686,7 +686,7 @@ private IPackageRepository CreateRepository(string repositoryUrl)
 				if (schemeName != null && Uri.CheckSchemeName(schemeName))
 					repositoryUrl = new Uri(repositoryUrl).AbsoluteUri;
 				else
-					repositoryUrl = new Uri("file:///" + Path.GetFullPath(Path.Combine(this.env.RootPath, repositoryUrl))).AbsolutePath;
+					repositoryUrl = Uri.UnescapeDataString(new Uri("file:///" + Path.GetFullPath(Path.Combine(this.env.RootPath, repositoryUrl))).AbsolutePath);
```
ilexp: Shouldn't the unescape be around the string being passed into the Uri in the first place, rather than the returned result from its `AbsolutePath` property?

Barsonax: Why is Uri actually used in the first place? Wouldn't Path.GetFullPath(Path.Combine(this.env.RootPath, repositoryUrl)) do the job already?

ilexp: > Putting it around the string that's passed into the Uri will make the bug reappear again. The Uri class itself will put %20 instead of spaces.

Confused unescape with escape for a second. Carry on!

> Why is Uri actually used in the first place? Wouldn't Path.GetFullPath(Path.Combine(this.env.RootPath, repositoryUrl)) do the job already?

Not sure to be honest, could be a remnant of an old copy-paste. Check all repository URL cases (web API url, file URL, local absolute path, local relative path) when changing this to see if they still behave as expected, just to be sure.

Barsonax: This line of code only seems to be used in the unit test as the 'if' makes sure that https will not hit that line. Path.GetFullPath will raise an exception if you give it a URL so the current version would crash too if it was fed a URL.

So that leaves only 'normal' paths which work fine.

ilexp: > This line of code only seems to be used in the unit test

Nope. You can specify any repository URL string in the config file, that's a supported use case, not just the unit test.

> Path.GetFullPath will raise an exception if you give it a URL so the current version would crash too if it was fed a URL.

That code path shouldn't be hit, as a file URL contains a scheme delimiter, which means it should prefer the first branch with a pure URL, not the second one. That's the intention anyway, so if you found that it doesn't actually work that way, then that's probably a bug.

Barsonax: No it does work. There is a if statement that makes sure it gets the correct strings (https strings go through another line of code)

PR #584, opened 2017-10-16
Fixed a bug in the rigidbody editor

ChristianGreiner:
When you select a rigidbody tool in the rigidbody editor view and deselect the current gameobject, the tool keeps selected. As soon as you click somewhere in the view, the editor throws an exception.

[Click for full size
](https://gfycat.com/LameHighGrayling)

So this bug is now fixed with this tiny pull request.


ilexp:
Merged. Thanks! 👍 

Hint for future commit messages: Make sure that your very first line is a short summary and only the following ones are hashtagged changes, as they won't be parsed properly in the first line otherwise.

Issue #491, opened 2017-01-30
Duality keeps recreating ProjectPlugins.


Barsonax:
### Summary

In my case i have renamed my solution to 'Pathfindax'. This works fine except for the fact that duality keeps recreating the ProjectPlugins solution file and the CorePlugin folder. From what I have seen in the source code the name of the solution file seems to be hardcoded.

### How to reproduce

-Rename solution file to a different name
-Restart duality
-Notice the ProjectPlugins solution file has been recreated.

### Workaround

Not really a workaround but you can ignore it in visual studio/source control and pretend its not there.

### Analysis

A way to change the default hardcoded name of the solution file would prevent this

ilexp:
It wouldn't hurt to allow changing this (similar to launcher selection), or wait with creating the solution until the user explicitly wants to open the game plugin source code.

Barsonax:
Another alternative would be that instead of searching for ProjectPlugins.sln it searches for *.sln inside the SourceCodeDirectory. This shouldn't be that hard to change I think and this will avoid recreating the ProjectPlugins.sln.

Eventually it would be nice if you could actually specify the name of the solution before they are added but this requires more work. After its created it could simply do the above check.

ilexp:
> Another alternative would be that instead of searching for ProjectPlugins.sln it searches for *.sln inside the SourceCodeDirectory. If this solution file exists there shouldn't be a reason to also generate the other projects.

Sounds good. If you're up for it, go ahead 👍

Barsonax:
Done :)

ilexp:
Fixed by @Barsonax in PR #585.

PR #585, opened 2017-10-17
#491

Barsonax:
See #491 

Moved all the project creating code inside the if(check if solution has to be created) block and changed the check to search for any .sln file


mfep:
I did a quick, mostly functional test on this. Here are my observations:
- Editor command to open source code **does not work**. It can be invoked by clicking the corresponding toolbar icon, or from the file menu. Cause: in [MainForm.cs](https://github.com/AdamsLair/duality/blob/a6156a851b1553825b6e13768b4fb2dde832f1fa/Source/Editor/DualityEditor/Forms/MainForm.cs#L569) it tries to open a hard-coded .sln path. Somehow `DualityEditorApp` should be asked about the current .sln filename.
- Other than that, functionality seems OK. Does not create additional solution files, and the namespace changing logic works too.
- Code style is OK.

The mentioned bug shouldn't be too difficult to fix, so I'd ask you, @Barsonax to add that. Thank you for the contribution!

Barsonax:
Oh wasnt even aware there is namespace changing logic (why?)?? Hmm does this also use hardcoded paths? Wil that still work if one changed the name of the csproj? 

Before this change the source button would have created a second solution and opened that. In that regard this is a improvement I guess XD.

Ill change it so that will work too now but it seems there are more hardcoded paths/logic than I first thought.

Barsonax:
Editor command to open source code will now work again. Added a property to DualityEditorApp called SolutionFileName which when called will:
-Check if the solution has been created already and if so return the path to the solution file
-If the solution is not created yet create it and return the path to the created solution file.

This does mean it will be searching for the solution file every time you call this property however as long as this doesn't happen a lot this shouldn't affect performance in a noticeable way. If this becomes a problem the path might have to be cached.

Another issue might be when the source folder contains more than 1 solution file.

I have been looking to the other code that modifies the csproj files and it seems to be doing all kind of things? Not really sure why this is needed as the template used to create the projects should contain the necessary relative paths already. If the user would change the name of the project this piece of code would be broken and I don't really like the idea of duality changing all kinds of settings in my csproj **after** its created. @mfep @ilexp

Source/Editor/DualityEditor/DualityEditorApp.cs
```diff
@@ -847,18 +850,44 @@ public static void ReadPluginSourceCodeContentData(out string rootNamespace, out
 				}
 			}
 		}
-		public static void InitPluginSourceCode()
+
+		/// <summary>
+		/// Initializes the solution unless there already is an sln file.
+		/// </summary>
+		/// <returns>The path to the solution file</returns>
+		public static string InitPluginSourceCode()
```
ilexp: After changing the solution file path property, the return value of this method is no longer used and it shouldn't really be used to retrieve information anyway - consider changing the return type to `void`.

Source/Editor/DualityEditor/DualityEditorApp.cs
```diff
@@ -847,18 +850,44 @@ public static void ReadPluginSourceCodeContentData(out string rootNamespace, out
...
 		{
-			// Create solution file if not existing yet
-			if (!File.Exists(EditorHelper.SourceCodeSolutionFile))
+			// Check if the solution file has to be created
+			string solution = FindSolutionFile();
```
ilexp: Suggestion: Rename to `existingSolutionPath` to clarify.

Source/Editor/DualityEditor/DualityEditorApp.cs
```diff
@@ -847,18 +850,44 @@ public static void ReadPluginSourceCodeContentData(out string rootNamespace, out
...
+				solution = FindSolutionFile();
+				// Shouldnt happen since the Properties.GeneralRes.GamePluginTemplate should always contain a solution file.
+				if (solution == null)
+					throw new FileNotFoundException("Could not find a solution file. The Properties.GeneralRes.GamePluginTemplate might have been changed and no longer contains a sln file.");
 			}
-			
+			return solution;
+		}
+
+		private static string FindSolutionFile()
```
ilexp: Suggestion: Consider merging the contents of this method with the updated `SolutionFilePath` property, since it won't be used anywhere else after the property change and is so small anyway. Otherwise, please move alongside the property into `EditorHelper` to keep them as close location-wise, as they're content-wise.

ilexp:
> I have been looking to the other code that modifies the csproj files and it seems to be doing all kind of things? Not really sure why this is needed as the template used to create the projects should contain the necessary relative paths already. 

They should, but last time I checked VS didn't support relative paths in that particular csproj property, so Duality patches those paths when needed. Moving the project shouldn't actually break this, as Duality should keep them updated accordingly - would be a bug if it didn't. 

It also updates them when selecting a different launcher application, which actually _is_ a case that can't be done statically in the template now that I think of it.

Barsonax:
Oh didnt knew you could switch out launchers. That would require patching the csproj. Doesn't sounds like this is a common use case though?

Still its only patching the projects using the hardcoded project paths. If the name does not match it will not find them. So this would be a potential bug.

However the problem then becomes figuring out what csproj to patch as simply patching them all would not work.

Source/Editor/DualityEditor/DualityEditorApp.cs
```diff
@@ -131,6 +130,10 @@ public static bool IsFirstEditorSession
 		{
 			get { return firstEditorSession; }
 		}
+		public static string SolutionFileName
+		{
+			get { return InitPluginSourceCode(); }
```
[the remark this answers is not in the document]
Barsonax: Agree, going a bit overboard here with the getter creating a entire solution :P

Barsonax:
Reverted the commit 4e34e602 and implemented this functionality more inline with the feedback from @ilexp.

Source/Editor/DualityEditor/DualityEditorApp.cs
```diff
@@ -849,116 +854,118 @@ public static void ReadPluginSourceCodeContentData(out string rootNamespace, out
...
+			// Check if the solution file has to be created		
+			if (EditorHelper.SourceCodeSolutionFilePath == null)
 			{
+				// Create solution file
 				using (MemoryStream gamePluginStream = new MemoryStream(Properties.GeneralRes.GamePluginTemplate))
 				using (ZipArchive gamePluginZip = new ZipArchive(gamePluginStream))
 				{
 					gamePluginZip.ExtractAll(EditorHelper.SourceCodeDirectory, false);
 				}
-			}
```
ilexp: Creating the solution file and initializing an existing solution file to match a project have been two independent operations, and they should remain independent for the time being, so the followup code should continue to be outside of the initial `if` block. It's possible that the current `Create Project` dialog relies on this.

Source/Editor/DualityEditor/Utility/EditorHelper.cs
```diff
@@ -40,6 +40,17 @@ public static class EditorHelper
...
 		private static VisualStudioEdition vsEdition;
 
+		/// <summary>
+		/// The path to the *.sln solution file. Will return null if the solution file has not yet been created.
+		/// </summary>
+		public static string SourceCodeSolutionFilePath
+		{
+			get
+			{
+				return Directory.EnumerateFiles(SourceCodeDirectory, "*.sln", SearchOption.AllDirectories).FirstOrDefault();
```
ilexp: Should not return null if no solution exists! It's just a path, no statement made about existence. Instead, return the default path at which the .sln will be created.

ilexp: In a second step, will need to replace the newly introduced `== null` checks elsewhere in code with file existence checks to match.

Source/Editor/DualityEditor/DualityEditorApp.cs
Barsonax: Yup my code broke the ```Create Project``` dialog. Will fix it.

PR #590, opened 2017-11-23
Cross Platform Path Combination

BobGneu:
Good Day, 

In light of the recent conversation and development towards cross platform support this PR gets ahead of the curve and ensures the paths are generated correctly, not depending on the `\` character which is windows specific. One side effect is that the strings that are generated must be `static readonly` and instead of _instanced constants_. 

Two further changes to discuss/consider: 
1. I would like to include removing the `@` verbatim strings, if that is something you are interested in. with the `\`s removed they are wholly unnecessary. 
2. It does not appear to be necessary to cache these values. We could convert them to Properties and pass them back in real time. 

Both changes are pretty straight forward, the later has more opportunity for impacting change across the framework/platform and so we may want to put that in a second PR, later. 

I have updated my game to run on this build, created a handful of new projects and did not run across any issues these changes would create. I did not see any test cases for this block and given that you would like them to be created I can put them together for you as well. 

I have synced up the formatting as best I could manage, I hope it complies. If not I will gladly correct any issues that present themselves. 

Best wishes.


ilexp:
Looks good to me. Assigning the community team for review to keep them in the loop, but I'd say this is a merge.

> It does not appear to be necessary to cache these values. We could convert them to Properties and pass them back in real time.

Don't see a reason why those values should change at runtime, so I'd keep them as static readonly fields for now.

> I would like to include removing the @ verbatim strings, if that is something you are interested in. with the \s removed they are wholly unnecessary.

Agreed, go ahead. Leaving the PR open for the removal commit, will merge afterwards.

Thanks for the PR!

ilexp:
Merged. Congrats to your first PR, and thanks for the contribution 🙂

Issue #150, opened 2014-12-09
Add a "Select Object" Button next to Reference Editors


AdamsLair:
- DragDrop is neat, but sometimes it just takes longer to search for the object in Scene or Project View than to simply open up a new window, enter a search term and select it.
- This button might replace the existing "highlight" button, because there's no space left. Maybe make the highlight button "optional", so that it only shows up when there's enough space?
- Open up a new window
  - Contains a flat list of applicable objects
  - Search field, focused by default
  - List is sorted by "match score" with the search field contents plus "most recently used score"
  - Hitting the return key on the reference property editor will open up the window. Hitting the return key in the window will select the topmost list item and accept it immediately. No mouse interaction required!
- Refactor the Create Object dialog to be actually a special case of this one, just with Types as "objects".

BobGneu:
Per the conversations with Adam and the crew in Discord, the request is to replace the highlight button here: 

![image](https://user-images.githubusercontent.com/857751/33594275-26174d88-d948-11e7-91ad-4f2e7936804e.png)

The replacement button should prompt the user with a filtered list of valid objects, allowing them to select from the elements in their project hierarchy. The selection box should 

* provide filters for relevance, recent usage. 
* It should only include relevant types in the listing. 
* should be the focus when the window is displayed
* should highlight the currently selected element when displayed. 
* should adhere to the color scheme of the application thus far
* should be modal

As it is, it looks like we have room for another button if we wish, so we can keep the highlight option if we want. 

As for an icon, we can use a funnel or similar to emphasize the filtering process. 

Due to the suggestion that we have a relevance, there may be a dependency to implement, although we can use things like word distance or a similar algorithm for the search implementation. In either case, configuration may be beneficial, even if not bubbled up to the user directly. I will note possible configuration options as this task comes together.

ilexp:
Very good analysis. Don't have much to add, just some ideas I got while reading your post:

> should be modal

Yep. Alternatively, should that feel too "limiting", we could explore a "close on lose focus" non-modal dialog type, but let's try the modal version first.

> As it is, it looks like we have room for another button if we wish, so we can keep the highlight option if we want.

Actually, I think we should remove the highlight button, but keep the functionality when clicking on the main / preview area of the reference editor. It is currently used for display only, but has no interaction aside from being a dragdrop receiver - might as well trigger a highlight when clicking it in order to clear some space for more important buttons.

> As for an icon, we can use a funnel or similar to emphasize the filtering process.

[...]

Issue #573, opened 2017-10-06
Switch From PCLs to .NET Standard


ilexp:
### Summary

Microsoft has officially deprecated Portable Class Libraries, which is what Duality is based on with regard to portability. Projects still compile just fine and will for the foreseeable future, but it will get progressively harder and less intuitive to keep things working and expanding on it. Consider switching all PCL projects to .NET Standard for v3.0.

### Analysis

- Duality PCLs use Profile111, which is equivalent to .NET Standard 1.1. Use this standard version for maximum compatibility.
  - Check which VS version is minimum required to create and build .NET Standard 1.1 libraries using the new reduced `csproj` format. Don't use the `.json` format, seems pretty much deprecated by now.
  - Be careful how NuGet packages are referenced. The `csproj` `PackageReference` item is not supported for all project types yet, see [here](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files). Might be better to stick to `packages.config` for now.
  - Consider doing a full switch to VS 2017 and the new project system as part of this change.
- .NET Standard packages require at least NuGet 2.12 or 3.4, so the Duality package manager will need to be updated to this first. The new `csproj` package reference format [requires at least NuGet 4.0](https://docs.microsoft.com/en-us/nuget/schema/project-json).
  - It may also need to handle platform-dependent dependency scenarios.
  - Make sure that Duality packages continue to be able to reference regular NuGet packages.
- Figure out how the NuGet package specs for Duality packages need to be changed in order in order to properly identify as netstandard packages.
- Get proficient with how .NET Standard works combined with regular .NET Framework applications and libraries when it comes to dependencies and runtime behavior.
  - Make sure the editor doesn't get spammed with .NET Framework specific forward-only assemblies.
  - Make sure the deployed launcher still works on platforms it worked before, like Linux and MacOS via Mono.
  - What's the role of reference assemblies in this?
- Some links for further research:
  - https://stackoverflow.com/questions/42254288/referencing-a-net-standard-library-from-a-windows-class-library
  - https://stackoverflow.com/questions/44617898/how-do-you-reference-a-net-standard-library-from-a-net-framework-4-5-console-a
  - https://github.com/dotnet/standard/issues/410
  - https://github.com/dotnet/standard/issues/481
  - https://www.hanselman.com/blog/ReferencingNETStandardAssembliesFromBothNETCoreAndNETFramework.aspx
  - https://github.com/dotnet/standard/issues/146#issuecomment-271105398
[...]

PR #593, opened 2017-12-14
DUALITY-150 - Resource Selection Dialog

BobGneu:
# Overview

This PR Provides the requested interface addition detailed in #150, to allow a user to select a resource from a filtered list. It provides all available resources to the user in a form that looks like the following: 

  

As can be seen in the screenshots, the header of the form changes to explain to the user the type of resource they are currently looking for. It filters from the list of resources provided by the `IContentProvider` interface. 

There are three major parts of this change proposal: 
1. A new form for filtered content selection
2. Updates to `ObjectRefPropertyEditor` interaction model:
    * Single clicking the resource readout panel now results in highlighting the currently referenced element in the project hierarchy
    * The button that was previously for highlighting now opens the filtered content form. 
3. Updates to the `ObjectRefPropertyEditor` derivative elements to track the type of the editor

I took a look through the test cases and I didn't see anything in the way of UI test cases for representative test cases, but I did run our current nunit test cases and they are all passing. I am quite interested in the thoughts you all have as to how you would like to proceed with automated test cases on a change of this nature. 

## Manual Test Cases & Validation

- [x] Single click on resource name readout panel and ensure resource is highlighted in the project tree view. 
- [x] Double click on resource name readout panel and ensure resource is highlighted in the project tree view.
- [x] Click button and see form
- [x] Form should list out all viable resources in the project, as returned by the `ContentProvider` interface
- [x] Entries should include content name as well as path to the item
- [x] Given that the user does not select anything, and they click `OK`  the resource reference should remain the same
- [x] Given that the user does not select anything and they click `Cancel` the resource reference should remain the same
- [x] Given that the user does not select anything and clicks the close button in the top right the resource reference should remain the same
- [x] Given that the user selects a new resource and clicks `OK` the resource reference should update to the new resource
- [x] Given that the user selects a new resource and clicks `Cancel` the resource reference should remain the same
- [x] Given that the user selects a new resource and clicks the close button in the top right the resource reference should remain the same
- [x] Given that the user opens the form, makes no change and then the application quits, the resource reference should remain the same
- [x] Given that the user opens the form and double clicks a resource entry the form should close and the resource reference should update to the new resource. 
- [x] Given that a `Component` references a `GameObject` it should provide a populated listing to select one from the current scene.
- [x] Given that a `Component` references a `Component` it should provide a populated listing to select one from the current scene.

I am open to further test cases. =)

## TODO

- [x] Document Public API
- [x] Correct tab indices to intuitive values
- [x] Currently selected item is selected when the form is initially shown 
- [x] Replace icon with new icon
- [x] Update `using` statements to be outside the namespace declarations, per S&P
- [x] Rename boolean from `buttonShowPressed` to `buttonSelectPressed`
- [x] `select` button should not be grayed out when no selection is present
- [x] Rename `ContentRefDialog` form to `ObjectRefSelectionDialog`
- [x] Rework form to not rely on the panels, and instead to leverage anchors. Form should leverage open space and draw from the `CreateObjectDialog` where possible.
- [x] Replace the ExplorerListView with TreeViewAdv.
- [x] Work with @ilexp to understand preferred method for assignment
- [x] Provide means of selecting GameObject, Components etc. _Include test cases for each when validating._
- [ ] Discuss Testing Options
- [ ] Work with team to select more fitting icon

## Further Enhancements

* We can change the double click behavior to result in selecting the currently referenced resource, but this functionality was not readily available and I wanted to make sure we followed the simple/small PR rule without too much scope creep. 
* Add preview of the resource to the content selection form. This also seemed to be a good addition, but the steps to implement it seemed to be beyond the obvious scope of the project. 
* Ignore replacing a reference with the same value as is currently referenced.


BobGneu:
Added icon as follows:

ilexp:
Thanks for the PR! I have assigned the team of community contributors for a first review session and will take a look myself as well soon.

SirePi:
Looks good to me. Well formatted and structured 👍

BobGneu:
Is there an icon you think wo old be more fitting?

ilexp:
Okay, so I did a quick code review (changes above) and also took a deeper look at behavior and implementation. Very clean PR so far, nicely minimal and to-the-point, big :+1: for that! 

We still have some things ahead before merging though, here's a list of things I found:

- The "select" button is displayed greyed out when the property value is null. This is a relic from its former times as a "show" button. Please adapt!
- There is currently no way to select other things than Resources, e.g. Components (deriving from / implementing Type X) and GameObjects. You already prepared for that in your addition of the `ReferenceType` property, what's missing is just the dialog support for these cases.
- Since the dialog will be used to select other things than Resources, it shouldn't be named after content ref selection - how about `ObjectRefSelectionDialog` to mirror the general `ObjectRefPropertyEditor` from which it originates?
- The `ExplorerListView` is currently used to display items, which is a control I'm trying to phase out. Can we use a `TreeViewAdv` instead? Shouldn't be much more code, has a lot more degrees of freedom with customization in mind, and it's overall a lot nicer to work with. Usage examples that can serve as a basis are the [CreateObjectDialog](https://github.com/AdamsLair/duality/blob/de927f8f15547e396ae6cb90c0f6ad3879621c3b/Source/Editor/DualityEditor/Forms/CreateObjectDialog.cs) and [WelcomeDialog](https://github.com/AdamsLair/duality/blob/de927f8f15547e396ae6cb90c0f6ad3879621c3b/Source/Editor/DualityEditor/Forms/WelcomeDialog.cs).
- There's a lot of complexity in the structure of the selection dialog, mostly due to the nested split panels. I don't think we really need them, as the same behavior can be achieved with regular WinForms anchors. I'd like to go a bit more into the direction of the CreateObjectDialog here, displayed below: 
https://i.imgur.com/8jt6dJz.png
https://i.imgur.com/C74BXAJ.png
- UI Design-wise, we need a bit of empty space on the edges to align with similar dialogs. Compared against the CreateObjectDialog:
https://i.imgur.com/9pKaq2E.png
https://i.imgur.com/layxmEw.png

As you will notice, some of them are high-level quality or consistency control, so we're on the right track here :) Let us know if you have questions or there's more to discuss. 

(Also, thanks to @AdamsLair/duality-contributors for the first level review - feel free to stay around and join the discussion!)

Source/Editor/DualityEditor/Forms/ContentSelectionDialog.cs
```diff
@@ -0,0 +1,82 @@
+﻿namespace Duality.Editor.Forms
+{
+	using System;
```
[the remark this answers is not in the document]
BobGneu: Noted. I'll update this.

Hopefully you are also aware of the [subtlety in resolution](https://stackoverflow.com/a/151560/16703) when the using statements are outside of the namespace declaration.

Source/Plugins/EditorBase/PropertyEditors/ObjectRefPropertyEditor.cs
```diff
@@ -18,9 +18,12 @@
 
 namespace Duality.Editor.Plugins.Base.PropertyEditors
 {
+	using Forms;
```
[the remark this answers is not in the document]
BobGneu: As above.

Source/Plugins/EditorBase/PropertyEditors/ObjectRefPropertyEditor.cs
```diff
@@ -322,26 +329,61 @@ protected override void OnMouseDown(MouseEventArgs e)
...
+				ResourcePath = this.ReferenceName
+			};
+
+			var result = tmpResourceSelectionForm.ShowDialog();
+
+			if (result == DialogResult.OK)
+			{
+				var tmpDataObject = new DataObject();
+
+				tmpResourceSelectionForm.SerializeToData(tmpDataObject);
```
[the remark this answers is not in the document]
BobGneu: I was hoping to keep this as minimal a change as possible, and given the way paste, copy, drag and drop are used I'm leery of deviating too far. Can you elaborate further on what you would like to see?

Source/Plugins/EditorBase/PropertyEditors/ObjectRefPropertyEditor.cs
```diff
@@ -322,26 +329,61 @@ protected override void OnMouseDown(MouseEventArgs e)
...
 			{
 				if (this.buttonResetPressed && this.buttonResetHovered) this.ResetReference();
 				this.buttonResetPressed = false;
-				this.Invalidate();
 			}
-			else if (this.buttonShowPressed && (e.Button & MouseButtons.Left) != MouseButtons.None)
+			else if (this.buttonShowPressed && leftMouseButtonUp)
 			{
-				if (this.buttonShowPressed && this.buttonShowHovered) this.ShowReferencedContent();
+				if (this.buttonShowPressed && this.buttonShowHovered)
```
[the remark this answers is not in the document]
BobGneu: Roger. This is where I thought refactoring out the input elements would be beneficial. For now I will focus on just addressing the comment, but ill make a note to come back to this later.

BobGneu:
Updated the initial 

Have you looked into using a [linter](https://www.sonarlint.org/visualstudio/index.html) or a service like [SonarQube](https://www.sonarqube.org/)/[SonarSource](https://www.sonarsource.com/) for quality controls and validation of the day to day contributions?

BobGneu:
@AdamsLair 

Would you like/mind if I remove the unused `using` statements?

Source/Plugins/EditorBase/PropertyEditors/ObjectRefPropertyEditor.cs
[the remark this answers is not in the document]
ilexp: Sure! So the reason for serialization at the point where I assume you copied it from is that we're stashing things in the Windows clipboard via Ctrl+C / Ctrl+V. Since this means that data can, in some situations, travel outside of Duality, serialization is a requirement to make this work.

With the new dialog, however, there is no need for data to ever leave Duality. Just hand it over like you would normally, e.g. ask the dialog what the user selected directly. One way to go about this would be to add a `SelectedResult` property to the dialog class, which is retrieved where you also check for the `DialogResult`.

ilexp:
> Would you like/mind if I remove the unused using statements?

I don't mind if you do, but it's not something I usually pursue either. Do as you see fit.

> Have you looked into using a linter or a service like SonarQube/SonarSource for quality controls and validation of the day to day contributions?

Nope, but as a first step, we have an [.editorconfig](https://github.com/AdamsLair/duality/blob/master/.editorconfig) for the rough C# style things, which is evaluated by the newer VS versions and in the long run I kind of want a set of [potentially Duality-specific analyzers](https://github.com/AdamsLair/duality/issues/298) to catch most of the style / convention things. This may conflict with [potentially switching to NetStandard](https://github.com/AdamsLair/duality/issues/573) due to the still incomplete tooling situation, but that's a topic for another day. In any case, automated checks that trigger while people are still working on the code are definitely something I want in the future. Hope it makes the contribution process easier for both sides.

> Updated the description of the PR to reflect requested changes. Will get into them as time permits.

Great! Ping me when you're ready for the next review stage 👍

BobGneu:
@ilexp 

> Sure! So the reason for serialization at the point where I assume you copied it from is that we're stashing things in the Windows clipboard via Ctrl+C / Ctrl+V. Since this means that data can, in some situations, travel outside of Duality, serialization is a requirement to make this work.
> 
> With the new dialog, however, there is no need for data to ever leave Duality. Just hand it over like you would normally, e.g. ask the dialog what the user selected directly. One way to go about this would be to add a SelectedResult property to the dialog class, which is retrieved where you also check for the DialogResult.

Earlier you suggested that I do a direct assignment instead of serializing. I have come about this a few ways and it appears that we are going to end up casting to a data object either way. Thoughts?

ilexp:
> Earlier you suggested that I do a direct assignment instead of serializing. I have come about this a few ways and it appears that we are going to end up casting to a data object either way. Thoughts?

So, the reason you need a data object is because you're implementing the dialog in the property editors base class - which doesn't really know the exact type or storage it is editing - instead of the derived classes, right? Because with a `DataObject` you can use the virtual deserialize method and each derived class does the more specific handling itself. This makes a lot of sense.

However! The dialog itself has no reason to assume it is used in the context of a property editor. It might as well be used in a custom UI, anywhere in the editor. It would make sense to keep it general-purpose and not specialize it towards this one use case, which happens to use a `DataObject`. So I think the solution that retrieves the result via regular property, and then wraps it in a data object for the property editor deserialize call is a good compromise.

Btw, great job on documenting your progress and the state of work. Very clean writeup on ToDos and requirements. 👍

BobGneu:
Are you suggesting we move the dialog into the instances of the editors? 

Each of them use the `DataObject` as a means of confirming assignment, and the serialization is incidental to the flow I am using. Even the function being referenced is only called serializing because it made sense to follow that pattern from the other code using `DataObject`s. Really all that was happening in that function was the reference path was being assigned to the `DataObject` to avoid having to pass an arbitrary `object` around and cast, and check again. As it is implemented now we leverage all of the work that have gone into the converter/`DataObject` flow.

ilexp:
> Are you suggesting we move the dialog into the instances of the editors?

Not sure what you mean by that, but I think the answer would be no 😄 I meant to reinforce that the dialog is in fact completely independent of the property editors and thus should _not_ have an API geared towards its use in this specific case - because it might as well be used elsewhere. 

This is the reason why I don't want the dialog itself to handle `DataObject` instances, but am suggesting to let the property editor acquire the dialog results and wrap them in a `DataObject` for its own use. As far as I can see, you're currently doing it this way.

BobGneu:
=) Alrighty. I think I follow your point in that case. 

I'll move over to provide the ability to set game objects now.

BobGneu:
@ilexp 

Added support for `Component` and `GameObject` selection, but i noticed an oddity I hadn't noticed previously. When selecting a component from the listing the `GameObject` name in the readout. Is that expected behavior?

ilexp:
@BobGneu If you're using the `ToString` method in order to retrieve the display name, then yes, that's perfectly normal. I would advise to use a custom method to derive names for GameObjects and Components instead, since `ToString` is more of a debug output and could change in the future.

BobGneu:
I didnt change the `DrawString` call, though I expect it may be using toString down the line. 
Going to take a deeper look.

BobGneu:
The `ComponentRefPropertyEditor` appears to want this behavior. It has already overridden the `ReferenceName` property to return the `GameObject` name if the `Component` has a `GameObject`. 

From `Source/Plugins/EditorBase/PropertyEditors/ComponentRefPropertyEditor.cs` [Line 28](https://github.com/AdamsLair/duality/blob/master/Source/Plugins/EditorBase/PropertyEditors/ComponentRefPropertyEditor.cs#L28)
```C#

		public override string ReferenceName
		{
			get { return this.component != null ? (this.component.GameObj != null ? this.component.GameObj.FullName : this.component.ToString()) : null; }
		}
```

Would you like for me to change this behavior?

ilexp:
Ohh, wait a second - I misunderstood your question. I thought you were talking about some default `TreeViewAdv` node name display. For the reference editors, displaying the GameObject name in place of the Component is expected behavior, since this describes the location of the referenced Component. No changes required here.

BobGneu:
Understood. 

Not to cause scope creep, considering this is new behavior and we may not have a great definition of expected behavior, but - under what contexts should we be able to gain access to a component for this reference tool that isn't also applied to a `GameObject`? Currently I am iterating over the scene for GameObjects with components based on the one referenced in the current component:

ilexp:
For GameObject selection, all GameObjects in the current scene are potential selection targets. For Component selection, you can use the results of `Scene.Current.FindComponents(type)` with the type being `EditedType` of the property editor. This will include all Components that either match the expected Component type exactly, or derive from it.

BobGneu:
Understood. I am currently using `Scene.Current.FindComponents` to get available components.

BobGneu:
I commented on the last post and it may have been overlooked.

> Also, in an effort to avoid using reflection I attempted to use the editedCmpType value, but given the nature of the value it causes the type to change if the reference is derived by others, e.g. with Component or a custom Camera component. I don't think this provides for a great user experience. Given the nature of the control we cannot rely on the current type assigned. I am thinking that reflection is the only real way forward here, but looking for your thoughts.

BobGneu:
In order to cache the value it looks like I would need to make a modification back in `AdamsLair.WinForms.PropertyEditing` which seems like it would be out of scope, considering it is in a dependency package. 

If you like I can fork the [other project](https://github.com/AdamsLair/winforms) and provide the patch [over there](https://github.com/AdamsLair/winforms/blob/master/WinForms/PropertyEditing/PropertyEditor.cs).

ilexp:
> Also, in an effort to avoid using reflection I attempted to use the editedCmpType value, but given the nature of the value it causes the type to change if the reference is derived by others, e.g. with Component or a custom Camera component. I don't think this provides for a great user experience. Given the nature of the control we cannot rely on the current type assigned. I am thinking that reflection is the only real way forward here, but looking for your thoughts.

The existing `EditedType` property of the `PropertyEditor` base class should provide the currently active property editor type, so no need to use reflection here. This type, however, dynamically adjusts depending on the currently selected value. This normally improves usability by allowing to edit more specific types, but in this case I completely agree that it makes more sense to stick to the original type.

Unfortunately, there is no easy, reliable way to achieve this through PropertyEditor API. The second best solution would probably be this:

1. Check if `EditedMember` is non-null. If it is, check if it is a `PropertyInfo` and retrieve its `PropertyType`. If it is not a `PropertyInfo`, check if it is a `FieldInfo` and retrieve its `FieldType`. If any type was retrieved this way, use this as a base type for the dialog.
2. If no type could be retrieved this way, use the type returned by `EditedType` instead.
3. If no type could be retrieved in any of the above ways, use `typeof(Component)`.

BobGneu:
Okay. Ill use those three, falling back each time. One side effect is that if you have a specific type you will have more elements in the listing than are relevant. The good news is that the converstion API already block the unwanted assignment for invalid values. 

Should have this patched in in a couple hours. Thank you for the feedback.

BobGneu:
Alrighty!

I think we have things squared away. Your fallback approach appears to work pretty well, I have added it as a protected property as follows: 

```C#
		protected Type FilteredType
		{
			get
			{
				Type t = null;

				if (this.EditedMember != null)
				{
					PropertyInfo tmpPropertyInfo = this.EditedMember as PropertyInfo;
					FieldInfo tmpFieldInfo = this.EditedMember as FieldInfo;
					if (tmpPropertyInfo != null)
					{
						t = tmpPropertyInfo.PropertyType;
					} else if (tmpFieldInfo != null) {
						t = tmpFieldInfo.FieldType;
					}
				}

				if (t == null)
				{
					t = this.EditedType != null ? this.EditedType : typeof(Component);
				}

				return t;
			}
		}
```

I am not fond of the name, but it mapped with the `FilteredType` property in the dialog. If you have a better name for the property we can make that change to finalize the changes. 

I have walked through all of the test cases discussed above and played around with the editor a bit more to see if I have covered my bases and it seems to be alright. This change appears to have been the final major code change. It appears to be the case that the pending notes are the icon and if you would like further energy spent in the automation front. I am going to look into removing the `+` from the tree and this branch should be good to go.

BobGneu:
One side note, for the second review: 
I added a new interface that the dialog implements at `Duality.Editor.Extensibility.DataConversion` called `IObjectRefHolder`. This interface simplifies the interface to assign to allow the implementation details to be driven by the `Component`, `GameObject` and `ContentRef` PropertyEditors, in turn. Not sure how you will feel about the use of this, but it seemed to be in line with our previous conversation regarding the abstract interface. 

I am not entirely comfortable with it being in the `DataConversion` namespace, but it seemed to be better suited there or as a struct in the `Duality.Editor.Plugins.Base.PropertyEditors`namespace where it would need to be setting all three properties blindly and then be passed into the `UpdateReference` method. I can switch either way, as you feel comfortable.

ilexp:
> I added a new interface that the dialog implements at Duality.Editor.Extensibility.DataConversion called IObjectRefHolder.

Looked into it quickly and I think the interface is unnecessary - so I'd prefer to do without it to keep things simpler. Here's two ways to achieve the same without the interface:

1. Write a bit of wrapper code in the `ObjectRefPropertyEditor` that handles the dialog, putting the result into the `DataObject` that is passed to deserialize. This way, you woudn't need the `UpdateReference` method either.
2. Change the `UpdateReference` to accept an `object` and adjust the callsite in `ObjectRefPropertyEditor` to put in the proper reference, which it can decide by simply checking which of those is set. To clean this up a bit and avoid having this code in the property editor, the dialog class itself could be adjusted to only have one result property of type `object` which carries whichever selection was made according to its filter type.

As a side note, if you _do_ have a virtual `UpdateReference` method, you don't really need the `DataObject` and deserialize call, as you have complete access to the selection field and type in the derived class where you override it. [This](https://github.com/BobGneu/duality/blob/7b23c30e2f8180e296fb87c3a364bd2c0015658e/Source/Plugins/EditorBase/PropertyEditors/ComponentRefPropertyEditor.cs#L113-L115) would be all the code you need in that case, in each of the three method overrides, adjusted to their derived editor.

> I think we have things squared away.

Great :) Let's address the above and then see if we can start to wrap things up with the next review stage.

BobGneu:
Sounds like a plan. Thanks for the feedback. 

I originally had the implementation seen here: bd01c503aaa4b62b413ff198c7c02bb027f9e831

```C# 
	if (tmpResourceSelectionForm.ResourceReference != null)
	{
		tmpDataObject.SetContentRefs(new[] {tmpResourceSelectionForm.ResourceReference});
	}
	else if (this.ReferenceType == typeof(GameObject))
	{
		tmpDataObject.SetGameObjectRefs(new[] {tmpResourceSelectionForm.GameObjectReference});
	}
	else
	{
		tmpDataObject.SetComponentRefs(new[] { tmpResourceSelectionForm.ComponentReference });
	}
```

Is this more in line with what you are looking for?

BobGneu:
Alrighty.
I think we are ready for the next stage.

ilexp:
> Is this more in line with what you are looking for?

Yep!

> Alrighty.
I think we are ready for the next stage.

Pinging @AdamsLair/duality-contributors for the next review, will join in later.

ilexp:
Approving my previous change requests, as they're outdated by now. Nice work so far 👍 

## ToDo (Functionality)

Checked the current version from the user side. Works as expected, didn't manage to break it or cause errors, so that's pretty good. However, found a few usability things that I think we should address before merging:



When the dialog is set to Resource mode, the fact that they're handled via `ContentRef<T>` should remain transparent, i.e. the content ref shouldn't be part of the window title. Instead, use the Resource type name itself, which you can obtain using a content refs [ResType](https://github.com/AdamsLair/duality/blob/de927f8f15547e396ae6cb90c0f6ad3879621c3b/Source/Core/Duality/ContentRef.cs#L62) property. 

Also, when displaying a type name, prefer [GetTypeCSCodeName](https://github.com/AdamsLair/duality/blob/de927f8f15547e396ae6cb90c0f6ad3879621c3b/Source/Core/Duality/Utility/ReflectionHelper.cs#L507) to stringify the type, as this will give you the name as it would be written in C# code.



Similar topic, when using a non-Resource mode, the dialog title shouldn't read `{type} Resource`, but simply `{type}`.



When displaying GameObject or Component nodes in the tree / list, the primary name should be the name of the GameObject, not the type name of the object.



Similarly, we could make better use of the path column in those cases by displaying the [full GameObject path](https://github.com/AdamsLair/duality/blob/de927f8f15547e396ae6cb90c0f6ad3879621c3b/Source/Core/Duality/GameObject.cs#L167).



Since we have two columns here, I think we should make proper use of the columns feature that TreeViewAdv provides, so we have a proper separation line between them and it looks a little tidier 🙂

Haven't looked into the code yet to see how this is currently done, but how I would approach it is roughly as follows:

1. Have separate NodeControl instances for the name and the path, each bound to a different Node property. You can probably just add another one in the WinForms designer and configure it. If not already existing, you might need to have a separate Name and Path property for your Node-derived class.
2. Enable the columns feature, should be a UseColumns property or similar.
3. Define the columns you want to use in the WinForms designer, one with the default icon / name NodeControls in it, the other one with the new path NodeControl in it.

That should be it. You might be able to skip some of the items, especially if you already have separate properties and NodeControls.

## ToDo (Implementation / Style / Wrap-Up)

Will look into this when we're done with the functionality side, but don't expect anything big to show up here. Expect tweaks and style things.

## Outlook

Here's some things that you could pull into the scope of this issue if you want, but that I would otherwise defer and put out there as a separate up-for-grabs issue.

- Removing the dotted line indicating a tree view, rather than a list view.
- Adding a search bar at the top of the dialog to filter the displayed nodes.
- Adding icons to the nodes.

BobGneu:
Thank you!

I'll address the feedback and add the first two of the **Outlook**  section in. I am not confident in getting the icon in there in a timely fashion. Should be able to work on this a bit tonight. 

Will update as soon as possible.

BobGneu:
Assignment broke for components so I pulled the type for the deserialization from the `FilteredType` field we created earlier.

BobGneu:
This should have addressed the comments about the title of the form and the `GameObject.FullName`

BobGneu:
Updated all references to the type to `GetTypeCSCodeName` when referencing type names

BobGneu:
New Layout using Columns

Here is the new layout with columns.

BobGneu:
New Layout using Columns

BobGneu:
Using a CueText field
And with an example
Here is the `CueTextBox` and an example of filtered input

BobGneu:
I think I have addressed the feedback, including the filter text field and column layout. 

Looking forward to your feedback in the morning.

ChristianGreiner:
I don't know if Duality overrides the WinForms-theme automatically.
In the project view you can see that the color of the columns of the list view looks different. (name / type)
image

I would use this style (flat borders) for the textbox (aaaand the buttons?):
image

BobGneu:
Just took a look at the other forms and it seems like there isn't a lot of standardization of the controls. 

New Project Dialog

BobGneu:
Here is the dialog with flat buttons

Flat Buttons on the Dialog

ilexp:
@ChristianGreiner Good call, the styling is indeed applied manually all over the editor.

> Here is the dialog with flat buttons

Duality just uses default buttons, so please do continue to use the default button style, no adjustments expected here. The screenshot @ChristianGreiner posted had flat buttons because that's probably Windows 10 :)

I agree on the other cases though: For text fields, please prefer single-line border and background color, for list columns, prefer a flat appearance. Adding a third point, I think the list view should keep the single-line border that was visible in your earlier screenshots. Looked quite nice!

The "New Project" dialog you both used as a comparison is in fact deviating from this, but you can check out the package manager dialog for a better example:



(Note: Button and combobox items are the default style on Win10)

------

Two more things I noticed on the list view: When I open the dialog, this is what I get initially:



After resizing, it snaps back to this:



Not sure what's going on here, but maybe something needs to be refreshed after populating the view initially, or maybe some initialization needs to take place in a different order?

Second, both the minimum and the starting width of the name column seems to be quite big - maybe reduce them by half?

## Summary

- Use default button style
- Flat appearance for column headers
- Single-line border and background color for search box
- Single-line border for list / tree view
- That weird layout bug when opening the dialog with a lot of content
- Minimum + Starting width of columns divided by 2

## Outlook

Updated with some new ideas. Again, 100% optional, would otherwise just put them in a new up for grabs issue after we conclude this PR.

- Adding icons for nodes.
- When opening the dialog, focus on the search field for direct keyboard input, auto-select the topmost list item and close the dialog on return, so it's 100% usable via keyboard.
- Hitting return when the property editor is selected should no longer invoke the "highlight object" action, but open the dialog.

BobGneu:
New Layout

BobGneu:
I am unable to reproduce the scrollbar situation in the latest build of the branch. I had already resized the dialog, but the behavior itself does not look to have been changed. 

### Attempted Repro Steps: 
1. Drop 50 Game Objects on the scene
2. Open the `ObjectRefSelectionDialog` and observe the layout before resizing
3. Resize the form
4. ensure that the scrollbars are always in the appropriate position on the TreeViewAdv control

### Repeat for 50 bitmaps

### Repeat for 50 Components

ilexp:
> I am unable to reproduce the scrollbar situation in the latest build of the branch. I had already resized the dialog, but the behavior itself does not look to have been changed.

Huh, weird. Thanks for checking! I'll take a look myself in the next review stage. See if I can find a fix or workaround with the repro on my machine.

Ping me when you're ready for the code review 👍

ChristianGreiner:
I think without the "Filter"-Label and the placeholder text in the textbox, it looks much "cleaner".
...like in your previous screenshot: 👌 
image

BobGneu:
Okay. I switched over to the label and text field due to the sample form @ilexp pointed to. I can switch back pretty quickly, so I'll hold off for a bit on any other feedback.

BobGneu:
I think I am ready for the next review. @ilexp

ilexp:
Added to my ToDo, will get back to you as soon as I manage 👍 

If any of the @AdamsLair/duality-contributors want to user test the latest changes or start with the code review, go ahead. I'll join in.

BobGneu:
@ilexp

What do you think about the cuetextbox vs. Text box and a label?

Source/Editor/DualityEditor/Forms/ObjectRefSelectionDialog.cs
```diff
@@ -0,0 +1,281 @@
...
+		{
+			ReferenceNode tmpNode = nodeAdv.Tag as ReferenceNode;
+			string tmpFilterValue = this.txtFilterInput.Text.ToLowerInvariant();
+
+			return tmpNode != null &&
+			    ( 
+					tmpNode.Name.ToLowerInvariant().Contains(tmpFilterValue) ||
+					tmpNode.Path.ToLowerInvariant().Contains(tmpFilterValue)
+				);
+		}
```
Barsonax: Minor thing but formatting seems to be a bit off

Source/Plugins/EditorBase/Properties/EditorBaseResCache.cs
```diff
@@ -14,6 +14,7 @@ public static class EditorBaseResCache
 	{
 		public static readonly Bitmap	DropdownSettingsBlack		= EditorBaseRes.DropdownSettingsBlack;
 		public static readonly Icon		IconEye						= EditorBaseRes.IconEye;
```
Barsonax: Formatting is a bit off here

Source/Plugins/EditorBase/PropertyEditors/ObjectRefPropertyEditor.cs
```diff
@@ -1,38 +1,36 @@
...
 		protected	Rectangle	rectPrevSound		= Rectangle.Empty;
 		protected	bool		buttonResetHovered	= false;
 		protected	bool		buttonResetPressed	= false;
-		protected	bool		buttonShowHovered	= false;
-		protected	bool		buttonShowPressed	= false;
+		protected	bool		buttonSelectHovered	= false;
+		protected	bool		buttonSelectPressed	= false;
 		protected	bool		panelHovered		= false;
 		protected	Point		panelDragBegin		= Point.Empty;
 		protected	Bitmap		prevImage			= null;
```
Barsonax: Formatting is a bit off here

Barsonax: Formatting is a bit off here

ilexp:
> What do you think about the cuetextbox vs. Text box and a label?

I don't have a strong opinion either way, maybe slight favor towards the cue text one, if any.

Source/Editor/DualityEditor/Forms/ObjectRefSelectionDialog.cs
```diff
@@ -0,0 +1,281 @@
...
+			}
+
+			public ReferenceNode(Component component) : base(component.GetType().GetTypeCSCodeName())
+			{
+				this.Name = component.GameObj.FullName;
+				this.Path = component.GetType().GetTypeCSCodeName();
+
+				this.ComponentReference = component;
+			}
+
```
ilexp: Code Style: 
* Fields should be next to fields, properties should be next to properties and methods should be next to methods. Do not mix them if it can be avoided within reason. 
* Keep statics at the top, instance methods below. 
* If the number of members of the same type grows beyond what can be easily observed with a full overview, try to group semantically where it makes sense. 
* Use two newlines to separate different member types, one newline to separate semantic groups within the same type, no newline between items of the same semantic group and type.

Example:
```csharp
public class Foo
{
	private int field1;
	private int field2;
	private int field3;
	
	
	public int Property1
	{
		get { /* ... */ }
		set { /* ... */ }
	}
	public int Property2
	{
		get { /* ... */ }
		set { /* ... */ }
	}
	public int Property3
	{
		get { /* ... */ }
		set { /* ... */ }
	}
	
	
	public Foo()
	{
		// ...
	}
	
	public void Method1()
	{
		// ...
	}
	public void Method2()
	{
		// ...
	}
	public void Method3()
	{
		// ...
	}
}
```

Source/Editor/DualityEditor/Forms/ObjectRefSelectionDialog.cs
```diff
@@ -0,0 +1,281 @@
...
+			private string _name;
```
ilexp: Code Style: Do not use underscore prefixes for fields or parameters.

Source/Editor/DualityEditor/Forms/ObjectRefSelectionDialog.cs
```diff
@@ -0,0 +1,281 @@
...
+		{
+			base.OnShown(e);
+
+			this.objectReferenceListing.ClearSelection();
+			this.objectReferenceListing.Model = this.Model;
+
+			this.objectReferenceListing.BeginUpdate();
+			this.Model.Nodes.Clear();
+			
+			if (this.FilteredType.IsSubclassOf(typeof(GameObject)) || this.FilteredType == typeof(GameObject))
```
ilexp: Suggestion: Use `typeof(GameObject).IsAssignableFrom(this.FilteredType)` to wrap this up in a single check. Also covers cases where you'd have an interface type instead of `GameObject`.

Source/Editor/DualityEditor/Forms/ObjectRefSelectionDialog.cs
```diff
@@ -0,0 +1,281 @@
...
+			{
+				this.Text = "Select a GameObject";
+
+				foreach (GameObject currentObject in Scene.Current.AllObjects)
+				{
+					ReferenceNode tmpNode = new ReferenceNode(currentObject);
+
+					this.Model.Nodes.Add(tmpNode);
```
ilexp: Suggestion: The newline between the two lines of the inner loop doesn't seem to be necessary. Keep it if you want (not a change request), but as a general rule of thumb, I try to make every newline express something meaningful and avoid those where it's not clear what exactly they're separating, like logical code blocks. Here, instantiating the node and adding the node each are so short and self-documenting that they might as well be regarded as a single logical unit, in which case they don't really need the newline.

Source/Editor/DualityEditor/Forms/ObjectRefSelectionDialog.cs
```diff
@@ -0,0 +1,281 @@
...
+				}
+			}
+			else if (this.FilteredType.IsSubclassOf(typeof(Component)) || this.FilteredType == typeof(Component))
+			{
+				this.Text = string.Format("Select a {0} Component", this.FilteredType.GetTypeCSCodeName());
+
+				foreach (Component currentComponent in Scene.Current.FindComponents(this.FilteredType))
+				{
+					ReferenceNode tmpNode = new ReferenceNode(currentComponent);
```
ilexp: Suggestion: This is also true for other places in code, but why is this variable prefixed with a `tmp`? It isn't really any more temporary than any other local variable, so might as well skip the prefix and give it a regular name. `node` would be just fine here as long as the context is clear and there's no ambiguity in scope.

Source/Plugins/EditorBase/Properties/EditorBaseResCache.cs
ilexp: This might actually be due to using tabs for alignment (as opposed to indentation) and GitHub using a different tab size. If it looks good in VS with a tab size of 4, that's alright. 

The proper solution would be to use spaces for alignment (but not indentation), which is a style guide that I only started implementing later in the project, so this file not doing it that way is me not having updated it yet. Since it has nothing to do with this PR, I'd just make sure the tab solution works in VS / with tab size 4 and not change anything else.

Source/Plugins/EditorBase/PropertyEditors/ObjectRefPropertyEditor.cs
ilexp: See tab alignment comment above.

Source/Plugins/EditorBase/PropertyEditors/ObjectRefPropertyEditor.cs
```diff
@@ -44,6 +42,37 @@ public abstract class ObjectRefPropertyEditor : PropertyEditor
...
+				Type t = null;
+
+				if (this.EditedMember != null)
+				{
+					PropertyInfo tmpPropertyInfo = this.EditedMember as PropertyInfo;
+					FieldInfo tmpFieldInfo = this.EditedMember as FieldInfo;
+					if (tmpPropertyInfo != null)
+					{
+						t = tmpPropertyInfo.PropertyType;
+					} else if (tmpFieldInfo != null) {
```
ilexp: Code Style: This should be
```csharp
}
else if (...)
{
```

Source/Plugins/EditorBase/PropertyEditors/ObjectRefPropertyEditor.cs
```diff
@@ -201,7 +230,7 @@ protected override void OnPaint(PaintEventArgs e)
 				rectText,
 				format);
 			
-			ControlRenderer.DrawBorder(e.Graphics, 
+			this.ControlRenderer.DrawBorder(e.Graphics,
```
ilexp: Good call adding the missing `this.` 👍

Source/Editor/DualityEditor/Forms/ObjectRefSelectionDialog.cs
```diff
@@ -0,0 +1,281 @@
...
+
+				foreach (IContentRef contentRef in ContentProvider.GetAvailableContent(tmpFilteredType))
+				{
+					ReferenceNode tmpNode = new ReferenceNode(contentRef);
+
+					this.Model.Nodes.Add(tmpNode);
+				}
+			}
+
+			this.objectReferenceListing.NodeFilter += this.NodeFilter;
```
ilexp: Why is this event subscription separate from the others in the constructor?

BobGneu:
Ill be going through the notes as soon as my munchkins are down. 

Thanks for the feedback @ilexp && @Barsonax

Source/Plugins/EditorBase/Properties/EditorBaseResCache.cs
BobGneu: Understood.

Source/Editor/DualityEditor/Forms/ObjectRefSelectionDialog.cs
BobGneu: Replaced the spaces with tabs.

BobGneu:
I think I have addressed the formatting issues and code quality issues raised in the earlier comments. 
I have also switched back over to the CueTextBox.

I think we are good to go. @ilexp

ilexp:
Sounds great, code looks good as well. Thank you for your work on this! I think I can take over from here with my side of the ToDo:

- Investigate and fix the window open layout bug that seems to be reproducible on my machine, but not elsewhere so far.
- Do a quick, final functionality test.
- Merge the PR into master.
- Trigger a binary release.
- Create a new issue for the outlook items that weren't pulled into this PR:
  - Adding icons for nodes.
  - When opening the dialog, focus on the search field for direct keyboard input, auto-select the topmost list item and close the dialog on return, so it's 100% usable via keyboard.
  - _I believe this was done already:_ Open dialog via return key on property editor

I'll post here with my progress on this.

ilexp:
Added some final touches to the code, merged to master and triggered a binary release. The new dialog will be out there in about 15 minutes.

I'll close issue #150 and proceed to create one for the remaining optional items.

ilexp:
Created new issues for the remaining items. 

We're done here. Thanks! 😃

Issue #595, opened 2017-12-25
Consider using matrices for transformation methods in the Transform component


Barsonax:
### Summary

See if its worth it to use matrices in the Transform component that can be reused across calculations as long as properties like position, scale and rotation do not change. The current idea is to add 2 matrices to the Transform component: 1 local and 1 world matrix. Since the Transform component is only interested in 2d 3x3 matrices would suffice. Once the matrix is built transforming a vector3 or vector2 will consists of only some multiplications and additions.

At this point its not clear yet what the best way is to integrate this into the Transform API. Should it be used under the hood or be exposed as a separate property for 'batch' operations? 

First focus on researching if its actually worth it and implement this as separate properties with no caching. The test code will handle the caching for now.

### Analysis
- Would avoid repeately calling expensive operations such as sin and cos. Most of the gains would be in batch operations. Though I suspect that if its reused even once it will outperform the current implementation.
- Would avoid having to cast to Vector3 when transforming a Vector2.
- Possibly cleanup the code a bit by moving some of the calculations out of the Transform component

ilexp:
Still skeptical as to whether this would pay off in the end when considering performance, but would be open to take a look at a prototype implementation and profiling results.

Some first thoughts:
* Public API should remain the same at first.
* There will need to be inverse matrices in addition to the regular ones to do inverse transform operations.
* Transformation is still a 3D operation - even though rotation only happens around the Z axis (2D), we do have 3D coordinates for translation. That means we need a 4x4 matrix for transformation.
* With between two and four 4x4 matrices added, memory footprint could become a concern. In addition to performance, memory size should be compared as well.

While a matrix-only implementation could be cleaner and more recognizable to contributors than the current specialized transformation code for Z rotation, scale and offset, that's still a lot of math and memory we don't actually need in our special case.

> See if its worth it to use matrices in the Transform component that can be reused across calculations as long as properties like position, scale and rotation do not change.

If caching is the main concern of this issue, we could actually just cache the `MathF.GetTransformDotVec` results for XY rotation + scale, which is essentially a 2x2 matrix and handle the translation separately. This wouldn't require restructuring all internals to use matrices and could be some middle ground. 

In this case too, memory footprint change and performance should still be evaluated.

Barsonax:
In a console application that referenced duality I managed to get a tenfold increase in performance for big batches when using a 4x4 matrix. The break even point is around 6 reuses.

[...]

PR #597, opened 2018-01-02
#595

Barsonax:


ilexp:
Will take a look at this later, but might take a bit longer than usually for a PR of this size. This is a `master` commit on the `Transform` component after all, so we need to make sure there's absolutely no added bug potential in there.

ilexp:
Simplified tests and Transform implementation a bit, made some minor code style adjustments, handled serialization and cloning for the new field and used it to optimize some more Transform methods and properties.

Still need to test some more real-world cases before merging.

Barsonax:
There is still room for optimization in UpdateAbs. It currently calls MathF.GetTransformDotVec which builds up a rotationmatrix and it also calls UpdateRotationDirAbs which updates the rotationDir. This is a duplicate sin/cos operation.

ilexp:
> There is still room for optimization in UpdateAbs. It currently calls MathF.GetTransformDotVec here which builds up a rotationmatrix and it also calls UpdateRotationDirAbs here which updates the rotationDir. This is a duplicate sin/cos operation I believe.

Nice find 👍 Will take care of it, still have the code open for testing anyway.

Issue #445, opened 2016-11-30
Introduce a way to load the last open scene in the editor


ChristianGreiner:
At the moment the editor doesn't open any scene after you launched it. 
It would be nice, if you could toggle an option, so the editor loads the last open scene automatically .

The toggleable option have to be stored in the AppData (default UserData).
(see [here](https://github.com/AdamsLair/duality/issues/13#issuecomment-137739542))

Some thoughts from the community:
> [22:20] SirePi: maybe a question, after you open "do you want to load the last edited scene?"
> [22:21] loeller: I think a popup is rather annoying
> [22:22] loeller: but most of the times you want to continue with the one you worked previously
> [22:23] loeller: actually, it can be a toggleable setting. But the popup at the start is not elegant imo

What do u think?

mfep:
'with the one you worked **on** previously'

ilexp:
Sounds good to me!

> The toggleable option have to be stored in the AppData (default UserData).
(see here)

AppData and UserData are for the game only, not the editor. This option would be part of the editor user settings, as serialized in `EditorUserData.xml`.

Since editor user data is not editable at a central place (as still planned in issue #13), we'd need to find a temporary place to put this option, or just default to true and require users who don't want this behavior to edit the `.xml` file by hand. The later option would be a good candidate for a first iteration.

Barsonax:
Made a PR that adds this functionality, still WIP.

ilexp:
Done by @Barsonax

PR #608, opened 2018-01-08
#445: duality now opens the last open scene

Barsonax:
Duality now opens the last open scene. TODO:
- [x] Add a way to disable this behavior
- [x] More testing

### DIFF
--- a/Source/Editor/DualityEditor/DualityEditorApp.cs
+++ b/Source/Editor/DualityEditor/DualityEditorApp.cs
@@ -52,7 +52,7 @@ namespace Duality.Editor
 		private	static GameObjectManager			editorObjects		= new GameObjectManager();
 		private	static HashSet<GameObject>			updateObjects		= new HashSet<GameObject>();
 		private	static bool							dualityAppSuspended	= true;
-        private	static List<Resource>				unsavedResources	= new List<Resource>();
+		private	static List<Resource>				unsavedResources	= new List<Resource>();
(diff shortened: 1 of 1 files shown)


Source/Editor/DualityEditor/DualityEditorApp.cs
```diff
@@ -63,6 +63,8 @@ public static class DualityEditorApp
 		private	static AutosaveFrequency			autosaveFrequency	= AutosaveFrequency.ThirtyMinutes;
 		private	static DateTime						autosaveLast		= DateTime.Now;
 		private	static string						launcherApp			= null;
+		private static string                       lastOpenScene       = null;
```
ilexp: Correct usage of spaces and tabs for alignment vs. indentation, but unfortunately the rest of the field definitions are not yet doing that, so they're not matching right now. Can you update the others too, so they match?

ilexp: Suggestion: The `lastOpenScene` could actually be a `ContentRef<Scene>`, which would make the fields usage a bit clearer, and also simplify the code you're using to switch if it exists.

Would merge either way, but putting this here in case you agree that this would be neat.

Barsonax: My untabify in VS is not working iam afraid :(

Barsonax: Done

ilexp:
Added some final touches and merged 👍 Thanks!

Issue #472, opened 2017-01-08
Introduce Serialization and Cloning Surrogates for HashSet<T>


ilexp:
### Summary

There already are serialization and cloning surrogates for `Dictionary<T,U>`, but there is none for `HashSet<T>` yet. While serialization and cloning still work, they're rather inefficient on `HashSet<T>` because of examining its internal structure as opposed to just dealing with the collection items. Introducing optimized surrogates for `HashSet<T>` would greatly improve this.

### Analysis

- The surrogate for `Dictionary<T,U>` can likely serve as a basis for the new `HashSet<T>` surrogates, though they will be a little more reflection-dependent because of the lack of a non-generic `ISet` interface.
- Introducing a new surrogate _should_ be backwards compatible, but double-check this.

Barsonax:
Opted to use the dynamic keyword since there is no non generic ISet interface. See PR #609.

Barsonax:
Came across a limitation of dynamic. If you put in a private type (like private class SomeClass {}) in the hashset the code will throw runtimebinding exceptions when calling a method like clear on the hashset. Its kinda weird but it seems like it threats the entire hashset as private a private type instead of just the contents. While this is a uncommon scenario it might happen. Still does it even make sense to serialize private types?

ilexp:
Continued this thread in PR #609, see [here](https://github.com/AdamsLair/duality/pull/609#issuecomment-358010097).

ilexp:
Done by @Barsonax.

PR #609, opened 2018-01-08
#472: added a hashsetsurrogate

Barsonax:
TODO:

- [x] Add  cloning unit tests
- [x] Add serialization unit tests
- [x] Rewrite cloning surrogate to use reflection
- [x] Rewrite serialization surrogate to use reflection
- [x] Write caching logic for the reflection code
- [x] Create a benchmark to measure performance
- [x] Optimize the reflection cache by moving most of the code to a generic method and only call this generic method with reflection


ilexp:
After adding a cloning surrogate, we'll also need a serialization surrogate to match. [Here's the one for dictionaries for example](https://github.com/AdamsLair/duality/blob/master/Source/Core/Duality/Serialization/Surrogates/DictionarySurrogate.cs), somewhat similar to the cloning ones.

> Unit tests?

Definitely! If we break this, a lot of projects are going to break as well. We'll need tests for this both on the serializer and the cloning side.

As far as serialization goes, there's an existing test for [HashSet<T> serialization](https://github.com/AdamsLair/duality/blob/9061c2b5da47e787a5ba1058907a821bd831e624/Test/Core/Serialization/SerializerTest.cs#L284), which you could extend to cover all the edge cases that a custom surrogate might stumble over:

- `null` values.
- Structs / non-primitive value types.
- Reference types.
- A list of two hashsets, each containing the same instances of some reference type. Check that their identities are preserved, so they're still the same instances after de/serialization.

As it turns out, there's also an [existing cloning test for hashsets](https://github.com/AdamsLair/duality/blob/9061c2b5da47e787a5ba1058907a821bd831e624/Test/Core/Cloning/SpecificCloningTest.cs#L51-L59) as well. Code looks very similar, and you can extend it to cover essentially the same edge cases.

No time for a proper review yet, but pinged the contributors group in case someone gets there faster than me. I'll get back to this later.

Source/Core/Duality/Cloning/Surrogates/HashSetSurrogate.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Duality.Cloning.Surrogates
+{
+	public class HashSetSurrogate : CloneSurrogate<dynamic>
```
ilexp: Since `dynamic` is a C# compiler alias, not an actual runtime type, please change this to `CloneSurrogate<object>` and use a local `dynamic` variable in the implementation to which the `object` parameter is assigned.

Source/Core/Duality/Serialization/Surrogates/HashSetSurrogate.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Duality.Serialization.Surrogates
+{
+	public class HashSetSurrogate : SerializeSurrogate<dynamic>
```
ilexp: See above.

Source/Core/Duality/Serialization/Surrogates/HashSetSurrogate.cs
```diff
@@ -0,0 +1,65 @@
...
+				Type hashSetType = hashSet.GetType();
+				Type[] genArgs = hashSetType.GenericTypeArguments;
+				TypeInfo keyTypeInfo = genArgs[0].GetTypeInfo();
+
+				hashSet.Clear();
+
+				object[] values;
+				reader.ReadValue("values", out values);
+				if (values != null)
+					foreach (dynamic obj in values)
```
ilexp: The `if` seems to be missing its brackets / block here.

Source/Core/Duality/Cloning/Surrogates/HashSetSurrogate.cs
```diff
@@ -0,0 +1,67 @@
...
+		{
+			Type hashSetType = source.GetType();
+			Type[] genArgs = hashSetType.GenericTypeArguments;
+			TypeInfo firstGenArgInfo = genArgs[0].GetTypeInfo();
+
+			if (!firstGenArgInfo.IsPlainOldData())
+			{
+				if (source == target)
+				{
+					foreach (dynamic key in source)
```
ilexp: Is there any reason to use `dynamic` for every item? Judging from the [dictionary surrogate](https://github.com/AdamsLair/duality/blob/master/Source/Core/Duality/Cloning/Surrogates/DictionarySurrogate.cs), `object` should be fine as well, and likely a lot faster.

Source/Core/Duality/Cloning/Surrogates/HashSetSurrogate.cs
```diff
@@ -0,0 +1,67 @@
...
+						setup.HandleObject(key, key);
+				}
+				else
+				{
+					foreach (dynamic key in source)
```
ilexp: See above

Source/Core/Duality/Cloning/Surrogates/HashSetSurrogate.cs
```diff
@@ -0,0 +1,67 @@
...
+
+			// Determine unwrapping behavior to provide faster / more optimized loops.
+			bool isKeyPlainOld = firstGenArgInfo.IsPlainOldData();
+
+			// Copy all values.
+			target.Clear();
+
+			if (!isKeyPlainOld)
+			{
+				foreach (dynamic key in source)
```
ilexp: See above.

Source/Core/Duality/Cloning/Surrogates/HashSetSurrogate.cs
```diff
@@ -0,0 +1,67 @@
...
+				{
+					dynamic keyTarget = null;
+					if (!operation.HandleObject(key, ref keyTarget)) continue;
+
+					target.Add(keyTarget);
+				}
+			}
+			else
+			{
+				foreach (dynamic key in source)
```
ilexp: See above.

Source/Core/Duality/Serialization/Surrogates/HashSetSurrogate.cs
ilexp: Also, use `object` for the item type unless there's a specific reason not to.

Source/Core/Duality/Cloning/Surrogates/HashSetSurrogate.cs
Barsonax: Yes if you use object you will get a runtimebinding exception as the method expects a T and you give it a object. Using dynamic will actually resolve this properly instead of giving a exception at runtime.

Source/Core/Duality/Cloning/Surrogates/HashSetSurrogate.cs
Barsonax: Yes if you use object you will get a runtimebinding exception as the method expects a T and you give it a object. Using dynamic will actually resolve this properly instead of giving a exception at runtime.

ilexp:
> Came across a limitation of dynamic. If you put in a private type (like private class SomeClass {}) in the hashset the code will throw runtimebinding exceptions when calling a method like clear on the hashset. Its kinda weird but it seems like it threats the entire hashset as a private type instead of just the contents. While this is a uncommon scenario it might happen. Still does it even make sense to serialize private types?
>
> Note: iam explicitly talking about a private type not a private field/property.

Can you clarify the case that is having problems with an example?

> Still does it even make sense to serialize private types?

Yes, this should be supported. It is up to a library / plugin author which classes to make public or private and nothing that should concern serialization or cloning support.

If this turns out to be a hard limitation of dynamic, then we need to adjust this PR to use reflection instead.

Barsonax:
Reproduced the dynamic issue in a console application:

```
class Program
{
	static void Main(string[] args)
	{
		HashSet<ReferenceType> source = new HashSet<ReferenceType>();
		ExtMethodsCloning.DeepClone(source);
	}

	private class ReferenceType		{	}
}

public static class ExtMethodsCloning
{
	public static void DeepClone(dynamic baseObj)
	{
		baseObj.Clear();
	}
}
```

If you put the DeepClone method inside the program class it works...It somehow needs to see ReferenceType or it will complain. Is this a bug in dynamic? This doesnt make much sense.

deanljohnson:
Possibly of interest for the dynamic problem: https://blogs.msdn.microsoft.com/davidebb/2010/01/18/use-c-4-0-dynamic-to-drastically-simplify-your-private-reflection-code/

Barsonax:
Its a different situation in that case though. Iam not calling any private methods. I call Clear on a HashSet<T> where the T is a private type not known to the code that calls it and it shouldnt need to as Clear doesnt care about the kind of T.

Still that solution might work as it removes the must be public limitation. Thats alot of code though just for this surrogate.

Barsonax:
Iam going to rewrite the code to use reflection instead of dynamic. I dont think its worth it to write a custom dynamic object just for the hashset surrogate. It was a fun experiment but it ends here :)

ilexp:
Sounds good 👍

Barsonax:
I think I can reduce the amount of reflection calls by moving most of the logic to a generic method and only use reflection to call this method. If this works thats just 1 reflection call instead of amountofelements + 1 calls. After that I could see if I can make that one call faster.

Barsonax:
Added hashset serialization tests

ilexp:
Neat! From what I can see, all ToDo checkboxes are ticked. If you're ready for review, I'll take a look soon.

Test/Core/Cloning/SpecificCloningTest.cs
```diff
@@ -23,7 +25,8 @@ namespace Duality.Tests.Cloning
 	[TestFixture]
 	public class SpecificCloningTest
 	{
-		[Test] public void CloneRegex()
+		[Test]
```
ilexp: Some auto-formatting went rogue in this file. Please revert the added newlines between the `[Test]` attributes and the test method. Regardless of whether or not this style is good, all the existing tests are written that way and we should remain consistent.

Test/Core/Cloning/SpecificCloningTest.cs
```diff
@@ -40,25 +43,53 @@ public class SpecificCloningTest
...
 		{
 			Random rnd = new Random();
-			HashSet<int> source = new HashSet<int>(Enumerable.Range(0, 50).Select(i => rnd.Next()));
+			var values = Enumerable.Range(0, 50).Select(i => rnd.Next());
+			HashSet<int> source = new HashSet<int>(values);
 			HashSet<int> target = source.DeepClone();
 
 			Assert.AreNotSame(source, target);
-			CollectionAssert.AreEquivalent(source, target);
+			CollectionAssert.AreEqual(source, target);
```
ilexp: `AreEquivalent` should be enough here. A hashset doesn't make any guarantees about element order and cloning a hashset would be allowed to change order, as long as the same elements are present.

Test/Core/Cloning/SpecificCloningTest.cs
```diff
@@ -40,25 +43,53 @@ public class SpecificCloningTest
...
+		[Test]
+		public void CloneHashSet_ReferenceType()
+		{
+			var rnd = new Random();
+			var values = Enumerable.Range(0, 50).Select(i => new TestObject(rnd, 0));
+			HashSet<TestObject> source = new HashSet<TestObject>(values);
+			HashSet<TestObject> target = source.DeepClone();
+
+			Assert.AreNotSame(source, target);
+			CollectionAssert.AreEqual(source, target);
```
ilexp: See above.

Test/Core/Cloning/SpecificCloningTest.cs
```diff
@@ -40,25 +43,53 @@ public class SpecificCloningTest
...
+		[Test]
+		public void CloneHashSet_ReferenceAndNulls()
+		{
+			var rnd = new Random();
+			var values = Enumerable.Range(0, 50).Select(i => i % 2 == 0 ? null : new TestObject(rnd, 0));
+			HashSet<TestObject> source = new HashSet<TestObject>(values);
+			HashSet<TestObject> target = source.DeepClone();
+
+			Assert.AreNotSame(source, target);
+			CollectionAssert.AreEqual(source, target);
```
ilexp: See above.

Test/Core/Cloning/SpecificCloningTest.cs
```diff
@@ -40,25 +43,53 @@ public class SpecificCloningTest
...
+		}
+
+		[Test]
+		public void CloneHashSet_ReferenceAndNulls()
+		{
+			var rnd = new Random();
+			var values = Enumerable.Range(0, 50).Select(i => i % 2 == 0 ? null : new TestObject(rnd, 0));
```
ilexp: This range contains 50% null values, but the hashset will only contain one. Maybe adjust the code to reflect that?

Test/Core/DualityTests.csproj.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
ilexp: This file shouldn't be checked in.

Test/Core/Serialization/SerializerTest.cs
```diff
@@ -281,12 +281,40 @@ public SerializerTest(Type format)
 			this.TestWriteRead(new CultureInfo("en-US"), this.PrimaryFormat);
 			this.TestWriteRead(new CultureInfo("en"), this.PrimaryFormat);
 		}
-		[Test] public void SerializeHashSet()
+
+		[Test]
```
ilexp: See rogue auto-format comment in the beginning.

Test/Core/Serialization/SerializerTest.cs
```diff
@@ -281,12 +281,40 @@ public SerializerTest(Type format)
...
+		public void SerializeHashSet_ValueType()
 		{
 			Random rnd = new Random();
 			HashSet<int> source = new HashSet<int>(this.CreateArray<int>(50, () => rnd.Next()));
 			HashSet<int> target = this.WriteRead(source, this.PrimaryFormat);
-			CollectionAssert.AreEquivalent(source, target);
+
+			Assert.AreNotSame(source, target);
+			CollectionAssert.AreEqual(source, target);
```
ilexp: See above.

Test/Core/Serialization/SerializerTest.cs
```diff
@@ -281,12 +281,40 @@ public SerializerTest(Type format)
...
+		[Test]
+		public void SerializeHashSet_ReferenceType()
+		{
+			var rnd = new Random();
+			var values = Enumerable.Range(0, 50).Select(i => new TestObject(rnd, 0));
+			HashSet<TestObject> source = new HashSet<TestObject>(values);
+			HashSet<TestObject> target = this.WriteRead(source, this.PrimaryFormat);
+
+			Assert.AreNotSame(source, target);
+			CollectionAssert.AreEqual(source, target);
```
ilexp: See above.

Test/Core/Serialization/SerializerTest.cs
```diff
@@ -281,12 +281,40 @@ public SerializerTest(Type format)
...
+		[Test]
+		public void SerializeHashSet_ReferenceAndNulls()
+		{
+			var rnd = new Random();
+			var values = Enumerable.Range(0, 50).Select(i => i % 2 == 0 ? null : new TestObject(rnd, 0));
+			HashSet<TestObject> source = new HashSet<TestObject>(values);
+			HashSet<TestObject> target = this.WriteRead(source, this.PrimaryFormat);
+
+			Assert.AreNotSame(source, target);
+			CollectionAssert.AreEqual(source, target);
```
ilexp: See above.

Test/Core/Serialization/SerializerTest.cs
```diff
@@ -281,12 +281,40 @@ public SerializerTest(Type format)
...
+		}
+
+		[Test]
+		public void SerializeHashSet_ReferenceAndNulls()
+		{
+			var rnd = new Random();
+			var values = Enumerable.Range(0, 50).Select(i => i % 2 == 0 ? null : new TestObject(rnd, 0));
```
ilexp: See above.

Test/Core/packages.config
```diff
@@ -2,4 +2,5 @@
 <packages>
   <package id="NUnit" version="2.6.4" targetFramework="net45" />
   <package id="NUnit.Runners" version="2.6.4" />
+  <package id="NUnitTestAdapter" version="2.1.1" targetFramework="net45" />
```
ilexp: This is already a global solution item, shouldn't be duplicated here.

Test/Core/DualityTests.csproj.DotSettings
Barsonax: Why not? It doesnt hurt ppl without resharper and it will only help (in this case) to keep the namespaces correct.

Test/Core/Cloning/SpecificCloningTest.cs
Barsonax: Any way to prevent VS from doing this?

Barsonax:
Changed the code according to the requests. Still not sure about csproj.DotSettings. This file contains resharper settings and was added since resharper was complaining about the namespaces not being the same as the folder structure. I set namespace provider to false for these folders so it ignores them when checking the namespace. It is not a file meant for local use.

Test/Core/DualityTests.csproj.DotSettings
ilexp: It's a third party tools settings file: Not helpful in regular VS, not required for the build, and whether or not someone uses Resharper is a matter of personal preferences. People who use that tool can still have that file locally.

ilexp:
> Still not sure about csproj.DotSettings. This file contains resharper settings and was added since resharper was complaining about the namespaces not being the same as the folder structure. I set namespace provider to false for these folders so it ignores them when checking the namespace. It is not a file meant just for local use but to let other ppl use it too.

Ah.. just read this after my earlier comment. Thanks for the detail!

> Well unless you don't want any resharper stuff in the repo.

For the moment, no. Just don't see Resharper has a strong claim for being part of the repo.

> Changed the code according to the requests.

Will take a look 👍

ilexp:
Merged. Thanks! 🙂

Issue #600, opened 2018-01-04
Investigate By-Ref Vector and Matrix Operators in C# 7.2


ilexp:
### Summary

As mentioned in issue #598, the new `in` parameter keyword is [allowed on operators and using literal values](https://docs.microsoft.com/en-us/dotnet/csharp/reference-semantics-with-value-types), so operator implementations can supersede the previous static by-ref operator alternatives.

### Analysis

- Wait until the core has adopted C# 7.2.
- Make sure using the `in` keyword implicitly with non-referencable values (like literals or property return values) has no negative impact on performance. Read the docs, and if that's not 100% clear, check out generated IL and x86 code in a sample project.
- Remove all static by-`ref` methods from `Vector2/3/4`, `Quaternion` and `Matrix3/4` if they have an operator equivalent.
- Use the `in` keyword on parameters provided to the operators in question.

Barsonax:
Also not sure if the current arithmetric implementation is the most optimal one. There is quite a big difference in the IL being generated. Take for example the vector3 + operator IL:
```
  .method public hidebysig static specialname valuetype Duality.Vector3 
    op_Addition(
      valuetype Duality.Vector3 left, 
      valuetype Duality.Vector3 right
    ) cil managed 
  {
    .maxstack 8

    // [502 4 - 502 22]
    IL_0000: ldarga.s     left
    IL_0002: ldflda       float32 Duality.Vector3::X
    IL_0007: dup          
    IL_0008: ldind.r4     
    IL_0009: ldarg.1      // right
    IL_000a: ldfld        float32 Duality.Vector3::X
    IL_000f: add          
    IL_0010: stind.r4     

    // [503 4 - 503 22]
    IL_0011: ldarga.s     left
    IL_0013: ldflda       float32 Duality.Vector3::Y
    IL_0018: dup          
    IL_0019: ldind.r4     
    IL_001a: ldarg.1      // right
    IL_001b: ldfld        float32 Duality.Vector3::Y
    IL_0020: add          
    IL_0021: stind.r4     

    // [504 4 - 504 22]
    IL_0022: ldarga.s     left
    IL_0024: ldflda       float32 Duality.Vector3::Z
    IL_0029: dup          
    IL_002a: ldind.r4     
    IL_002b: ldarg.1      // right
    IL_002c: ldfld        float32 Duality.Vector3::Z
    IL_0031: add          
    IL_0032: stind.r4     

    // [505 4 - 505 16]
[...]

PR #610, opened 2018-01-10
#600 Vector and point operator micro optimizations

Barsonax:


BobGneu:
What were the performance improvements to this optimization?

SirePi:
@BobGneu from what I understood in the discussions in the chat, this brings a major speed increase in the execution of the altered operations; I'm just wondering if and how it does affect memory and GC performances due to the high amount of allocations that this change implies, since Vector operations are *a lot* in every update cycle. Maybe I'm just paranoid though 😅

Barsonax:
Well first of all these are structs that are allocated on the stack so I don't think we have anything to do with GC at all here.

Basically the compiler did not optimize the code as far as we would like. We helped the compiler a bit by rewriting the code.

Iam no expert at x86 however but if you look at the assembly it is quite clear that its faster (google the instruction names). See issue #600 for source, IL and assembly code. The old situation was using some pointer like stuff I believe (lea instructions?) and now its all very straight forward.

SirePi:
Ah, perfect then 👍

Source/Core/Primitives/Math/Point2.cs
```diff
@@ -102,104 +102,79 @@ public static float Distance(Point2 left, Point2 right)
 		/// </summary>
 		public static Point2 operator +(Point2 left, Point2 right)
 		{
-			left.X += right.X;
-			left.Y += right.Y;
-			return left;
+			return new Point2(left.X + right.X, left.Y + right.Y);
```
ilexp: Can you change the style of the constructor calls in this PR into a multi-line format? Especially with repeating patterns, I think this makes it a lot more readable and easier to spot typos. Consider this:

```csharp
return new Vector4(
	left.X + right.X,
	left.Y + right.Y,
	left.Z + right.Z,
	left.W + right.W);
```

Barsonax: Done

ilexp:
Merged!

PR #611, opened 2018-01-11
Unicode font importer

SirePi:
The new FontAssetImporter allows to import unicode blocks and characters in a more simple way than the current extendedCharSet version, at the cost of possibly importing more characters than what is actually needed.

**ExtendedCharSet** (taking in input a string made of the characters that are desired) has been replaced by the **UnicodeBlocks** list, initialized by default with _BasicLatin_ and _Latin-1 Supplement_ Unicode Blocks (chars 00 to ff), that allows the user to specify the Unicode Blocks to import, and text area **CustomCharSet** that takes in input a string akin to MS Word's printing page selector, with hex codes of the desired characters and ranges (i.e. **0-7f,80-ff,200** will import characters from _0_ to _127_, _128_ to _255_, and _512_); specific error messages will illustrate eventual issues in the text input by the user.
Control characters are excluded from the import phase.


SirePi:
Potential drawback: the user now needs to know the Unicode Block or Unicode code of the characters that he is interested in importing (nothing that a visit to https://unicode-table.com cannot fix). Has ExtendedCharSet ever been used before? Is that even an issue?

SirePi:
Being an action that is performed only during import, I didn't care that much about performances as opposed as to make the code simple to read.
One thing that I was thinking about though is, now that it's so easy to import a lot of characters, might there be an issue of creating a texture that is simply too big to manage by itself? (i.e. CJK ideographs are a single block of 22000 glyphs.. if you want to have them at 50x50 pixels, a 8192 texture is just barely enough). Again, is this an issue? maybe on older video cards? Should the importer be able to generate a number of bitmap planes and manage them in a single resource? Just ignore this for now and come back if the issue actually arises?

Barsonax:
> Being an action that is performed only during import, I didn't care that much about performances as opposed as to make the code simple to read.

Then it doesnt really matter.

> One thing that I was thinking about though is, now that it's so easy to import a lot of characters, might there be an issue of creating a texture that is simply too big to manage by itself? (i.e. CJK ideographs are a single block of 22000 glyphs.. if you want to have them at 50x50 pixels, a 8192 texture is just barely enough). Again, is this an issue? maybe on older video cards? Should the importer be able to generate a number of bitmap planes and manage them in a single resource? Just ignore this for now and come back if the issue actually arises?

Generating a single 8k texture could pose a problem on some systems. It might be better to split I guess. However as long as this is not a regression I think a clear error message is fine for now.

ilexp:
> Generating a single 8k texture could pose a problem on some systems. It might be better to split I guess. However as long as this is not a regression I think a clear warning/error message is fine for now.

Definitely not a regression, this limitation has always been there. Would see gracefully handling this as further improvement, not requirement for this PR. Providing a proper error message would be something to be done on the graphics backend side (throw a specialized exception when exceeding max texture size), so a different issue as well.

> Just ignore this for now and come back if the issue actually arises?

Sounds good 👍 

----

Review is still on my ToDo, will get back to you when I can.

Issue #606, opened 2018-01-07
Add Icons to Object Reference Selection Dialog


ilexp:
### Summary

The [Object Reference Selection Dialog](https://github.com/AdamsLair/duality/commit/bf11b157d5d0edaccc57862fbc447e4d6352e0de) displays a filterable list of GameObjects, Components or Resources from which the user can select a single item. Readability could be improved by adding the familiar icons to those items, so they can be easily recognized.

### Analysis

- To enable icon support in the `TreeViewAdv` UI element, a `NodeStateIcon` needs to be added as a first item before the existing two `NodeControl` items. It will display whatever icon is assigned to each nodes `Image` property.
- Next, each `ReferenceNode` constructor will need to set this property. This could be done using the `GetEditorImage` extension method on the specific objects type of each node.

ilexp:
Done by @deanljohnson. Closing this.

PR #615, opened 2018-01-15
Object ref dialog icons

deanljohnson:
### Summary
Implemented #606. The implementation follows almost exactly as described in the issue text. There is one decision I made that is debatable, however.

The ReferenceNode class has three Reference properties that had public get/setters. With the edits I made, the Image property is assigned within the inspector based on the value of the Reference property passed into the constructor. If these properties were left with public setters, some later code could set these Reference properties to a new object that could need a different image than the one originally set in the constructor. The simplest solution was to make the setters private. Now, future uses of this class avoid a possible bug.

The alternative is to change the getter/setters to use private backing fields and to update the Image whenever the a Reference property changes. This would certainly work, but just seemed like more work than is necessary right now. If requested, I can implement that.

### DIFF
--- a/Source/Editor/DualityEditor/Forms/ObjectRefSelectionDialog.Designer.cs
+++ b/Source/Editor/DualityEditor/Forms/ObjectRefSelectionDialog.Designer.cs
@@ -32,151 +32,172 @@ namespace Duality.Editor.Forms
 		/// </summary>
 		private void InitializeComponent()
 		{
-			this.buttonCancel = new System.Windows.Forms.Button();
-			this.buttonOk = new System.Windows.Forms.Button();
-			this.labelInfo = new System.Windows.Forms.Label();
-			this.objectReferenceListing = new Aga.Controls.Tree.TreeViewAdv();
-			this.columnName = new Aga.Controls.Tree.TreeColumn();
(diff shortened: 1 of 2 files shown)


Source/Editor/DualityEditor/Forms/ObjectRefSelectionDialog.Designer.cs
```diff
@@ -105,13 +109,22 @@ private void InitializeComponent()
...
+			this.columnImage.TooltipText = null;
+			this.columnImage.Width = 32;
+			// 
 			// columnName
 			// 
 			this.columnName.Header = "Name";
 			this.columnName.MaxColumnWidth = 300;
 			this.columnName.MinColumnWidth = 80;
 			this.columnName.Sortable = true;
-			this.columnName.SortOrder = System.Windows.Forms.SortOrder.Ascending;
```
ilexp: Why was sorting by name disabled?

deanljohnson:
It shouldn’t have been - my mistake. I’m not in a place where I can fix it
right now, but will update the PR later.

deanljohnson:
Fixed the name column sorting.

ilexp:
Merged. Thanks! 🙂

PR #616, opened 2018-01-15
More text goodies

SirePi:
Couple of simple improvements:

- Added unicode character capability to **FormattedText**
- otf (OpenType Font) import suport for **FontAssetImporter**

### DIFF
--- a/Source/Core/Duality/Drawing/FormattedText.cs
+++ b/Source/Core/Duality/Drawing/FormattedText.cs
@@ -53,6 +53,10 @@ namespace Duality.Drawing
 		/// </summary>
 		public	const	string	FormatNewline		= "/n";
 		/// <summary>
+		/// Format string for inserting an unicode character.
+		/// </summary>
+		public	const	string	FormatUnicode		= "/u";		
+		/// <summary>
 		/// Returns a format string for changing the current text color to the specified one.
 		/// </summary>
 		/// <returns></returns>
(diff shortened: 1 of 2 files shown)


ilexp:
Quick question about the new `/u` FormattedText control char: Since C# string already support unicode, why wouldn't you just use a unicode character directly?

SirePi:
Mostly because the editor (or I should say the editor's font) does not support unicode so it displays them all as squares; after having to import (and use) 20 or so non-latin characters, it became difficult to understand what was written where

ilexp:
Hmm.. can we remove this from the PR? I agree that the editor situation you described is not great, but I would prefer to address the actual issue (Object Inspector fonts for text property editors), rather than add a new feature to the core that the core itself doesn't actually need.

As a workaround until then, would it work if you copy-paste the text you're editing to external text editor of your choice for affected texts? I usually do this for long texts where the current single-line text editor isn't great either, could work in your case as well.

SirePi:
Sure, go ahead 😄 just comment out/remove lines 874-889?

ilexp:
Removing the lines in question sounds good. You or me, whoever get back here first 👍

SirePi:
Done

SirePi:
Well... thinking back about it, there is the case where you are using a font that includes characters in the unicode private range, and as such has no representation, or simply do not correspond to the "standard" meaning of the character.. Of course it's still possible to add the character by hand, or by copying it in the editor and have it displayed as a square, but this would not be solved by an ad-hoc font, which would have the side effect of actually masking the situation. Having the ability to actually type your desired code would help in this kind of situation.

As I said before, it's more like a facility for in-editor editing, doing it programmatically you can always compose your string as you better desire.

Your choice 😄

ilexp:
You have a point, I'm just not sure this is the "right" way to address the issue - and I think it could pay off to wait with this until things get a bit clearer because more experiences have been made.

ilexp:
And it's merged. Thanks! 🙂

PR #622, opened 2018-01-31
Added undo history module

padesso:
A basic editor plugin that binds the UndoRedoManager.ActionStack to a ListBox.  Compliant with coding styles and no API docs required.  Test functionality as well as look and feel.


ilexp:
There seem to be build errors in this PR? Also, given its "new feature" nature, a more thorough description of what this PR does would be appreciated 🙂

padesso:
Sorry about that - not sure why it failed.  I integrated it into a fresh copy and will look at it soon to make sure the automated build doesn't fail and provide a more detailed description.

padesso:
Fixed the build issue with a bad reference in my plugin project to AdamsLair.Winforms and WeifenLuo.WinFormsUI.Docking by mimicking reference path in other editor plugin project.

This editor plugin is implemented in a consistent fashion to other editor plugins with regard to project structure and code locations.  The UI component, UndoHistoryView, extends DockContent and can be accessed using the View menu item within the editor.  Once opened, it can be docked similarly to other windows.

The UI contains a ListView that is bound to the UndoRedoManager.ActionStack.  Performing an action in the editor modifies the UndoRedoManager.ActionStack and is thus reflected in the UI.  Performing an undo/redo action using shortcut keys or the Edit menu behaves as before but you may also undo or redo by clicking on a entry in the UndoHistoryView UI.

Source/Plugins/EditorModules/UndoHistoryView/Modules/UndoHistoryView.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Forms;
+using WeifenLuo.WinFormsUI.Docking;
+
+namespace Duality.Editor.Plugins.UndoHistoryView
+{
+    public partial class UndoHistoryView : DockContent
+    {      
+        private BindingSource _undoStackSource;
```
deanljohnson: Should be named "undoStackSource"

Source/Plugins/EditorModules/UndoHistoryView/Modules/UndoHistoryView.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        public UndoHistoryView()
+        {
+            InitializeComponent();
+
+            //Can't bind directly to a IReadOnlyList, so we create a binding source
+            _undoStackSource = new BindingSource();
```
deanljohnson: Access to the ""undoStackSource" member should use "this.undoStackSource"

Source/Plugins/EditorModules/UndoHistoryView/UndoHistoryViewPlugin.cs
```diff
@@ -0,0 +1,73 @@
...
+using Duality.Editor.Properties;
+using System;
+using System.Windows.Forms;
+using WeifenLuo.WinFormsUI.Docking;
+
+namespace Duality.Editor.Plugins.UndoHistoryView
+{
+    public class UndoHistoryViewPlugin : EditorPlugin
+    {        
+        UndoHistoryView _undoHistoryView;
```
deanljohnson: Should be named "undoHistoryView"

ilexp:
Didn't manage to do a review yet, but so far I agree with the points @deanljohnson mentioned. I'll add a work-in-progress tag on this PR for addressing them. Let us know when you're ready for the next iteration, or in case you have questions along the way.

padesso:
Thanks guys. I'm on the road but will get it wrapped up this week.

I don't necessarily agree with the second point:


   - Clicking on an item in this list automatically undoes and redoes
   actions to move the stack point to that location. Undo/Redo should only
   occur by clicking one of the arrows.

I think this is a workflow that users will appreciate. I do.

ilexp:
> I think this is a workflow that users will appreciate. I do.

You may have a point that people might appreciate fewer clicks to achieve the same, and I guess there isn't really a reason to select any item from the list, other than to jump to that point in history. Let's try it your way then, and see how it goes.

ilexp:
@padesso Any update on this?

padesso:
Apologies, not yet. Going to out some time in this weekend to hopefully
address the button states.

ilexp:
Okay, so I finally found the time to look at this from a user perspective, but I don't think we should merge this in its current state. A typical use case for an Undo / Redo history will be that it remains open while the user works on the project. This, however, is currently not possible because the history window consumes so much time redrawing itself while editing that there is no longer a good user experience. Take a look at [this screencap](https://gfycat.com/DefiantWigglyIbis) for reference. You can also see unnecessary redraws when jumping to a point in history in [this screencap](https://gfycat.com/SlimyPalatableEmu).

Two things need to change:

- **Redrawing itself needs to be faster**, and also flicker-free. I'm suspecting that ListBox simply suffers from the typical issues that are common to almost all native WinForms collection Controls and that the easiest way to address and fix this would be to use TreeViewAdv instead. It's used for lists, trees and tables all over the editor, is flicker-free and has never suffered from performance issues so far. It uses a model / view approach and I suspect that it should be possible to implement an "UndoRedoStackModel" that sits directly on the action stack.
- **There needs to be a better redraw strategy**. This may involve adding a new event to the UndoRedoManager API that indicates a change in the currently active action vs. a change to the stack as a whole. That way, continuous editing would be distinguishable from starting a new edit operation, allowing to greatly reduce the number of updates required in the history UI.

Some pointers on the TreeViewAdv control:

- Custom tree model examples:
  - [Type Hierarchy](https://github.com/AdamsLair/duality/tree/master/Source/Editor/DualityEditor/Controls/TreeModels/TypeHierarchy)
  - [Folder Browser](https://github.com/AdamsLair/duality/tree/master/Source/Editor/DualityEditor/Controls/TreeModels/FileSystem)
- TreeViewAdv usage examples:
  - [WelcomeDialog](https://github.com/AdamsLair/duality/blob/master/Source/Editor/DualityEditor/Forms/WelcomeDialog.cs)
  - [CreateObjectDialog](https://github.com/AdamsLair/duality/blob/master/Source/Editor/DualityEditor/Forms/CreateObjectDialog.cs)
  - [ObjectRefSelectionDialog](https://github.com/AdamsLair/duality/blob/master/Source/Editor/DualityEditor/Forms/ObjectRefSelectionDialog.cs)

padesso:
I haven't been able to carve out time to get started on the latest
recommendation.  I will close the PR and re-open down the road when I can
dedicate time to it.

Issue #426, opened 2016-10-29
Object Inspector Undo / Redo Support for Adding / Removing Elements to IList and IDictionary


ilexp:
### Summary

While the Object Inspector allows you to Undo / Redo modifying list and dictionary elements properly, there is currently no implementation for Undo / Redo support on adding and removing elements.
### Analysis
- There is no Add / Remove element event handler in list and dictionary property editors. This would be a required change in `AdamsLair.WinForms`.
- There is no Undo / Redo action that represents this kind of operation in a generic way yet.

deanljohnson:
I'm thinking about looking into this. When I first started contributing to Duality, you weren't accepting PR's for `AdamsLair.WinForms`. Is that still the case?

ilexp:
Nope, all the Duality-related projects are open for PRs now. If you want to tackle this, go ahead!

deanljohnson:
Awesome. Here are some notes on my progress and what I have learned:

I was first going to take an event based approach where the IList and IDictionary PropertyEditors send events whenever an item is added to a list they control. This works fine on the `AdamsLair.WinForms` side of things, but I had trouble figuring out the right way to hook into these events from Duality.

As far as I can tell, the DualityPropertyGrid does not have a point I could hook into for when a property editor is removed. So while I could subscribe to the added and removed events on the IList and IDictionary editors, I couldn't unsubscribe. Furthermore, creating an UndoRedoAction in this design is a problem. Other UndoRedoActions contain the code that does and undoes their associated actions. With the event design I was planning, the action is already done within the PropertyEditors.

Now I'm thinking about taking a similar approach to the [IList index setters](https://github.com/AdamsLair/winforms/blob/c59be26c15ced5ed713e43f3b4db9d4360ca8e51/WinForms/PropertyEditing/PropertyEditors/IListPropertyEditor.cs#L30) and [IDictionary key value setters](https://github.com/AdamsLair/winforms/blob/c59be26c15ced5ed713e43f3b4db9d4360ca8e51/WinForms/PropertyEditing/PropertyEditors/IDictionaryPropertyEditor.cs#L25) where they use a delegate to perform the operation on the collections they control. I'd write a default that adds and removes items as the editors do now, but allow the delegate to be switched to something on the Duality side.

[...]

PR #623, opened 2018-02-01
List dict add remove undo support

deanljohnson:
### Summary

Implemented Undo and Redo support for IList and IDictionary property editors as discussed in #426. Writeback is automatically handled through the macro system of the UndoRedoManager. 

There exists a minor issue with the AddKeyToDictionary action. Undo'ing the action requires two ctrl+z's. It seems like typing in the `Add Entry` field is pushing an `EditPropertyAction` onto the undo/redo stack. I'm not sure how to remove that or how to append it to the AddKeyToDictionary action. CanAppend is not called on the AddKeyToDictionary action.

Another minor issue is the use of `DualityEditorApp.NotifyObjPropChanged(this, new ObjectSelection(Scene.Current));` to allow the PropertyEditors to update after the action is done/undone. We don't have references to the target objects and so we can't send a more specific `ObjPropChanged` signal. This is used in quite a few other UndoRedoActions so I don't think it's that bad, but it'd be nice to be able to emit a more specific signal.

### DIFF
--- a/Build/NuGetPackageSpecs/Base.Editor.nuspec
+++ b/Build/NuGetPackageSpecs/Base.Editor.nuspec
@@ -2,7 +2,7 @@
 <package>
   <metadata>
     <id>AdamsLair.Duality.Editor.Plugins.Base</id>
-    <version>2.2.11</version>
+    <version>2.2.12</version>
     <authors>Fedja Adam</authors>
     <owners>Fedja Adam</owners>
     <title>Editor Base Infrastructure</title>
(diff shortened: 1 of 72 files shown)


ilexp:
cc @AdamsLair/duality-contributors for first-level code review.

ilexp:
> There exists a minor issue with the AddKeyToDictionary action. Undo'ing the action requires two ctrl+z's. It seems like typing in the Add Entry field is pushing an EditPropertyAction onto the undo/redo stack. 

That's weird. Not strictly speaking part of the issue or PR, but since you're on similar code, can you check out why that happens / where it is triggered? This probably shouldn't trigger an UndoRedo in the first place. If there's an easy fix, great. Otherwise, the double Ctrl+Z could just remain for now, to be solved later.

> Another minor issue is the use of DualityEditorApp.NotifyObjPropChanged(this, new ObjectSelection(Scene.Current)); to allow the PropertyEditors to update after the action is done/undone. We don't have references to the target objects and so we can't send a more specific ObjPropChanged signal. This is used in quite a few other UndoRedoActions so I don't think it's that bad, but it'd be nice to be able to emit a more specific signal.

This should be totally fine. The editor wide `NotifyObjPropChanged` is actually intended for objects that are edited directly, i.e. ones that you can make a selection of an edit in the Object Inspector or a specialized editor. More fine-grained notifications should be mapped to their next biggest parent, because in the end someone else needs to determine whether a change event affects their UI, and they don't know or want to deal with any specific objects current implementation details.

ilexp:
Merged!

Issue #550, opened 2017-07-15
Simple Math and Geometry convenience improvements


mfep:
### Summary
The following suggestions were gathered while using the engine for a number of projects. The intention is to make one's life easier by providing shortcuts for common calculations.

### List of suggestions

1. Vector component properties for `Vector2` and `Vector3` types. Obtaining the X, Y and Z vector components quickly. Names are arguable.
Implementation example:
``` csharp
struct Vector3 {
    Vector3 XVector { get { return new Vector3 (this.X, 0f, 0f); } }
    Vector3 YVector { get { return new Vector3 (0f, this.Y, 0f); } }
    Vector3 ZVector { get { return new Vector3 (0f, 0f, this.Z); } }
}
```

2. More coordinate selector properties for `Rect`. Convenient access for: `TopMiddle`, `RightMiddle`, `BottomMiddle` and `LeftMiddle` coordinates.

3. A saturation function to the `MathF` class. Similar to `Sign`, but accepts an epsilon value too.
Implementation example:
``` csharp
float Saturate (float val, float eps)
{
    if (Abs (val) <= eps) return 0f;
    return Sign (val);
}
```
4. `Clamp` is quite often used with `0f` and `1f` as boundaries. Let's create a function for that: `Clamp01`.

5. Inverse linear interpolation to `MathF`, and possibly to `Vector2` and `Vector3`.
Implementation example:
``` csharp
float InvLerp (float value, float min, float max)
{
    return (value - min) / (max - min);
}
```

6. Inverse linear interpolation with `0f` and `1f` as boundaries.

7. Introduce the well known `smoothstep` function to `MathF`. The formula is `x * x * (3 - 2 * x)`.

ilexp:
> 1. Vector component properties for Vector2 and Vector3 types. Obtaining the X, Y and Z vector components quickly. Names are arguable.
Implementation example:

> 3. A saturation function to the MathF class. Similar to Sign, but accepts an epsilon value too.
Implementation example:

I think they're a bit too specific, so I'd leave them out for now.

> 2. More coordinate selector properties for Rect. Convenient access for: TopMiddle, RightMiddle, BottomMiddle and LeftMiddle coordinates.

Sounds reasonable. I don't want to introduce "middle" as a new terminology to the rect class, but I we could probably go with either "center" instead, or nothing at all, e.g. `Top`, `Right`, `Bottom` and `Left` each returning the middle of that side as a Vector2. What do you think?

[...]

PR #625, opened 2018-02-14
Added Rect properties.

superstewie:
Added new Rect properties: Top, Bottom, Left, Right, as outlined in Issue #550 .
Could not resolve NUnit missing reference in a timely manner, so created local tests.
    - These tests were not committed.


superstewie:
Disregard this pull request. I forgot to handle the case where we have a negative width/height, and also a corresponding negative coordinate.

Issue #612, opened 2018-01-13
Extract flood fill algorithm from FillTilemapTool


ChristianGreiner:
At the moment the [flood fill alogrithm](https://github.com/AdamsLair/duality/blob/7febba7196982e9fa7903a501378bf6e121157fe/Source/Plugins/Tilemaps/Editor/CamViewStates/TilemapTools/FillTilemapTool.cs#L121-L316) is implemented directly in the fill tilemap tool. 

It would be clever if this alogrithm would be extracted, because you could reuse this code for other tilemap tools like a magic wand selection tool.

ilexp:
Sounds good. Would be important to make sure that we don't introduce any performance downgrade by abstracting too much, since the flood fill is actually a bottleneck for very big maps. Do you have a design proposal for this?

Barsonax:
Is it just me or does the flood fill looks overly complex? Shouldnt it be a simple BFS algorithm?

ChristianGreiner:
> Do you have a design proposal for this?

Maybe we could create a new class called `FloodFillArea` inside the [Utility](https://github.com/AdamsLair/duality/tree/7febba7196982e9fa7903a501378bf6e121157fe/Source/Plugins/Tilemaps/Editor/Utility) folder and move all methods and stuff from the FillTilemapTool into this class. 

Any other suggestions?

ilexp:
The Utility folder you linked is part of the editor plugin like the current place of the algorithm. I think this will be more useful if we find a core plugin location for the algorithm, similar to [TilemapCulling](https://github.com/AdamsLair/duality/blob/7febba7196982e9fa7903a501378bf6e121157fe/Source/Plugins/Tilemaps/Core/Tilemaps/TilemapCulling.cs). Maybe call the class TilemapFloodFill and put it next to it?

deanljohnson:
This one is done, right? (https://github.com/AdamsLair/duality/pull/627)

ilexp:
It has. Thanks! Closing this.

PR #627, opened 2018-02-28
Extract flood fill algorithm from FillTilemapTool

ChristianGreiner:
https://github.com/AdamsLair/duality/issues/612

### DIFF
--- a/Source/Plugins/Tilemaps/Core/Tilemaps/TilemapFloodFill.cs
+++ b/Source/Plugins/Tilemaps/Core/Tilemaps/TilemapFloodFill.cs
@@ -18,12 +18,12 @@ namespace Duality.Plugins.Tilemaps
 		/// <param name="floodFillArea"></param>
 		/// <param name="floodFillOrigin"></param>
 		/// <returns>True, if the algorithm completed. False, if it was canceled.</returns>
-		public bool GetFloodFillArea(Tilemap tilemap, Point2 pos, bool preview, Grid<bool> floodFillArea, ref Point2 floodFillOrigin)
+		public bool GetFillArea(Tilemap tilemap, Point2 pos, bool preview, Grid<bool> floodFillArea, ref Point2 floodFillOrigin)
 		{
 			Grid<Tile> tiles = tilemap.BeginUpdateTiles();
 			Point2 fillTopLeft;
 			Point2 fillSize;
-			bool success = FloodFillTiles(ref this.activeFillBuffer, tiles, pos, preview ? (128 * 128) : 0, out fillTopLeft, out fillSize);
+			bool success = FillArea(ref this.activeFillBuffer, tiles, pos, preview ? (128 * 128) : 0, out fillTopLeft, out fillSize);
 			tilemap.EndUpdateTiles(0, 0, 0, 0);
 
 			// Find the filled areas boundaries and copy it to the active area
@@ -38,8 +38,8 @@ namespace Duality.Plugins.Tilemaps
 		}
 
 		/// <summary>
-		/// Performs a flood fill operation originating from the specified position. 
-		/// <see cref="Tile"/> equality is checked in the <see cref="_FloodFill_TilesEqual"/> method.
+		/// Performs a flood fill operation originating from the specified position.
+		/// <see cref="Tile"/> equality is checked in the <see cref="FillAreaTilesEqual"/> method.
 		/// </summary>
 		/// <param name="fillBuffer">A buffer that will be filled with the result of the flood fill operation.</param>
 		/// <param name="tiles"></param>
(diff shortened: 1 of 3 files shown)


Barsonax:
I myself am not a big fan of static state due to its potential for all kind of nasty side effects (also it keeps it in memory forever unless you null it). It should be possible to modify the code so that static is no longer needed.

Aside from that looks clean.

Source/Plugins/Tilemaps/Core/Tilemaps/TilemapFloodFill.cs
```diff
@@ -0,0 +1,213 @@
+﻿using System;
+
+namespace Duality.Plugins.Tilemaps
+{
+	/// <summary>
+	/// A helper class that provides a flood fill algorithm.
+	/// </summary>
+	public static class TilemapFloodFill
```
Barsonax: We are dealing with static state here now due to the static activeFillBuffer field. Remove static from this class and its members and just use a instance of the TilemapFloodFill in FillTilemapTool to solve this.

Source/Plugins/Tilemaps/Core/Tilemaps/TilemapFloodFill.cs
```diff
@@ -0,0 +1,213 @@
...
+	{
+		private static Grid<bool> activeFillBuffer = new Grid<bool>();
```
Barsonax: We are dealing with static state here now due to the static activeFillBuffer field. Remove static from this class and its members and just use a instance of the TilemapFloodFill in FillTilemapTool to solve this.

ChristianGreiner: Yeah, you are right. I tought that this class should be static too.. (see TilemapCulling -> https://github.com/AdamsLair/duality/blob/master/Source/Plugins/Tilemaps/Core/Tilemaps/TilemapCulling.cs).

ilexp: I think I can see the arguments on both sides here. It makes sense to keep it static for consistency reasons, but it also makes sense not to, for the reasons @Barsonax mentioned. I'm personally leaning towards the instance variant, now that I've seen it mentioned. 

One request I'd add is to make sure the same instance of the flood fill algorithm is re-usable for multiple operations, re-using its internal buffers along the way to avoid allocations.

Barsonax: FillTilemapTool is reused for multiple operations right? Then making the buffer a instance field of FillTilemapTool will make it reuse the buffer for multiple operations. If not then it never reused it in the first place before these changes.

ilexp: Yep, the tools are kept alive and re-used. Even if they weren't though, I think an "algorithm class" instance should be re-usable by design whenever possible.

Barsonax:
Excuse me I meant making the buffer a instance field of TilemapFloodFill. My mistake.

ChristianGreiner:
I made some changes (see last commit) @Barsonax @ilexp

ChristianGreiner:
@ilexp Done 👍 Do you agree with the changes?

Source/Plugins/Tilemaps/Core/Tilemaps/TilemapFloodFill.cs
```diff
@@ -0,0 +1,213 @@
...
+			// Get the base tile for comparison
+			Tile baseTile = tiles[pos.X, pos.Y];
+
+			// Find the topleft-most tile to start with
+			fillAreaTopLeft = FillAreaFindTopLeft(fillBuffer, tiles, pos, baseTile);
+			fillAreaSize = new Point2(1 + pos.X - fillAreaTopLeft.X, 1 + pos.Y - fillAreaTopLeft.Y);
+			pos = fillAreaTopLeft;
+
+			// Run the main part of the algorithm
+			if (maxTileCount <= 0) maxTileCount = int.MaxValue;
```
BobGneu: Given that this we are already leaving off the `{}` on single lined if blocks this should either be moved down and tabbed up or a space should follow so the assignment doesn't seem related.

Source/Plugins/Tilemaps/Core/Tilemaps/TilemapFloodFill.cs
```diff
@@ -0,0 +1,213 @@
...
+			}
+			while (lastRowLength != 0 && pos.Y < tiles.Height);
+
+			return true;
+		}
+
+		private Point2 FillAreaFindTopLeft(Grid<bool> fillBuffer, Grid<Tile> tiles, Point2 pos, Tile baseTile)
+		{
+			// Find the topleft-most connected matching tile
+			while (true)
```
BobGneu: given that we have a test for completion shouldn't we rework this to have the test be up here?

ilexp:
@BobGneu Since this PR is primarily about extracting the algorithm implementation, I'd say all improvements to the algorithm itself are entirely optional. Merging now, but further improvements are of course appreciated!

PR #629, opened 2018-03-06
Default graphics backend now follows init rules

JerTH:
Fixed a bug where DefaultOpenTKBackendPlugin would initialize OpenTK even if Duality did not select it as the graphics backend


ilexp:
Merged. Thanks! 🙂

Issue #634, opened 2018-03-19
Remove the solution level packages


Barsonax:
### Summary
Solution level packages have been deprecated here: https://github.com/NuGet/Home/issues/522#issuecomment-98845250

Currently the packages.config doesnt do anything can can be removed.

ilexp:
Done by @Barsonax in PR #635. Closing this.

PR #635, opened 2018-03-19
#634: removing solution level packages

Barsonax:

### DIFF
--- a/Duality.sln
+++ b/Duality.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.40629.0
+# Visual Studio 15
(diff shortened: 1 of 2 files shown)


Duality.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.40629.0
+# Visual Studio 15
+VisualStudioVersion = 15.0.27428.1
```
ilexp: Technically not a clean PR with this change, as updating the VS version of the solution could backfire and wasn't part of the issue - but I'll take this as it make sense to update the solution version by now. Would prefer a more selective PR next time though 🙂

ilexp:
Merged. Thanks!

PR #643, opened 2018-04-22
Made flat perspective scroll more uniform

TheBoJaN:
Zooming in flat mode is buggy and inconsistent due to FocusDistIncrement, multiplying Delta by current FocusDist is much smoother and provides a more natural zoom experience.

### DIFF
--- a/Source/Plugins/EditorModules/CamView/CamViewStates/CamViewState.cs
+++ b/Source/Plugins/EditorModules/CamView/CamViewStates/CamViewState.cs
@@ -812,7 +812,7 @@ namespace Duality.Editor.Plugins.CamView.CamViewStates
 				}
 				else
 				{
-					this.View.FocusDist = this.View.FocusDist + this.View.FocusDistIncrement * e.Delta / 40;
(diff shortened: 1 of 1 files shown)


ilexp:
Hey there and thanks for your contribution!

@AdamsLair/duality-contributors Nothing much to review here code-wise, but maybe quickly check behavior of this change vs. current release for cases like:

- Flat / Perspective projection in Cam View
- Camera near / behind zero-Z plane
- Regular wheel zoom and holding-spacebar-down zoom

Would be willing to merge if you give a thumbs up.

TheBoJaN:
Good point, that can easily fixed by changing it to
`this.View.FocusDist += ((float)e.Delta / 1000f) * Math.Max(this.View.FocusDist, 0.1f);`
This ensures you can always scroll back from 0 RefDist. I've also tested to make sure it doesn't have any adverse effects with cameras at different Z positions.

This way of zooming makes it much easier to achieve high precision zoom without it drastically over-adjusting. I've also made some other changes in my personal build to improve panning at different zoom levels but I may need to test it out some more.

ilexp:
Package release in progress. Should be out there in about 20 minutes.

Issue #644, opened 2018-04-24
Dualitor camera zooms uncontrollably when attempting to zoom in/out and move camera simultaneously


babyccino:
### Summary

In the camera view in Dualitor, if you zoom in or out using mouse scroll while moving the camera laterally using the middle mouse click, the camera will continue zooming in or out until you release middle mouse. The zooming in/out also gets increasingly fast the longer middle mouse is held.

### How to reproduce

- In Dualitor open a camera view
- Hold middle mouse
- While still holding middle mouse scroll the mouse wheel

PR #645, opened 2018-04-24
Fix issue with scroll when camera is being moved

babyccino:
Fix for issue #644 by decreasing the camera's z velocity when the camera is moving.


ilexp:
Merged. Thanks!

ilexp:
Package release in progress. Should be out there in about 20 minutes.

Issue #637, opened 2018-03-21
Introduce a Way to search for a Component before adding it to a GameObject.


ChristianGreiner:
At the moment is quite annoying when you want to add a new component via the context menu,
especially when you have a "big project structure"...

Here is an example:
![image](https://user-images.githubusercontent.com/6233308/37707780-9c38935e-2d04-11e8-8d4b-51ae8a28e9c0.png)

So it would be nice if you could search for a specific component instead of clicking through the context menu.
Here is an example:
![image](https://user-images.githubusercontent.com/6233308/37707922-1c0b87c6-2d05-11e8-940d-4824c98e451f.png)
_by Unity_

We could reuse this dialog by @BobGneu for this propose.
![image](https://user-images.githubusercontent.com/6233308/37708146-dd16e8e8-2d05-11e8-9044-eaaa7caa4297.png)
 

So what do you say?

ilexp:
Looks like an overdue usability improvement 👍 Let's say we go with the pop-up dialog variant, how would you trigger the dialog?

ChristianGreiner:
One way to trigger the dialog could be a button on the right hand side (red button)?

![image](https://user-images.githubusercontent.com/6233308/37724901-111b34c4-2d32-11e8-8770-0257731090ca.png)

mfep:
Just a little addition: I'd be happy to see that in the *Add new GameObject* context menu as well

ilexp:
> One way to trigger the dialog could be a button on the right hand side (red button)?

Sounds good. We could extend this even further:

- A small (but maybe not red?) button to the top right of the GameObject property editor in the inspector. Not super prominent, but always reachable. Can be implemented in the [GameObjectPropertyEditor](https://github.com/AdamsLair/duality/blob/c260bba93d739cfb25723f8aeb624792bc4a6a80/Source/Plugins/EditorBase/PropertyEditors/GameObjectPropertyEditor.cs).
[...]

PR #648, opened 2018-05-01
#637 Add Component Dialog

deanljohnson:
### Summary

Added support for a component search dialog as discussed in #637. The dialog is accessible from the SceneView and from the Object Inspector.

### Details
* Used the previously named ObjectRefSelectionDialog to provide the selection
* Renamed ObjectRefSelectionDialog to SelectionDialog to reflect its more varied use cases
* See image for location and look of buttons added to the Object Inspector

image

image


ilexp:
Already looks pretty good 👍 Putting this out for review.

ilexp:
Code looks good at a glance, will give it a quick behavior / usability check later and get back to you.

Source/Editor/DualityEditor/DualityEditor.csproj
```diff
@@ -121,11 +121,11 @@
     <Compile Include="Extensibility\Previews\PreviewQuery.cs" />
     <Compile Include="Extensibility\Previews\IPreviewQuery.cs" />
     <Compile Include="Extensibility\Previews\PreviewSizeMode.cs" />
-    <Compile Include="Forms\ObjectRefSelectionDialog.cs">
+    <Compile Include="Forms\SelectionDialog.cs">
```
ilexp: Agree 100% to renaming the dialog, but `SelectionDialog` sounds a bit too generic. Not sure what a better name would be, but maybe something like "ObjectSelectionDialog" (without the ref) or "SearchableListSelectionDialog"? Open for suggestions.

Source/Plugins/EditorBase/PropertyEditors/AddComponentPropertyEditor.cs
```diff
@@ -0,0 +1,113 @@
...
+using Duality.Editor.UndoRedoActions;
+using ButtonState = AdamsLair.WinForms.Drawing.ButtonState;
+
+namespace Duality.Editor.Plugins.Base.PropertyEditors
+{
+	/// <summary>
+	/// A <see cref="PropertyEditor"/> that provides a button
+	/// for adding a component to a <see cref="GameObject"/>
+	/// </summary>
+	public class AddComponentPropertyEditor : PropertyEditor
```
ilexp: Works as expected, but for improved usability, the editor should also react to the return key in order to "press" the button - that way the Object Inspector remains usable completely via keyboard. There should be a virtual method you can override to react to keyboard input while the editor is selected.

Source/Plugins/EditorBase/PropertyEditors/AddComponentPropertyEditor.cs
```diff
@@ -0,0 +1,113 @@
...
+			this.Height += 4;
+		}
+
+		protected override void UpdateGeometry()
+		{
+			base.UpdateGeometry();
+
+			this.rectButton = new Rectangle(
+				this.ClientRectangle.X + 2, this.ClientRectangle.Y + 2,
+				this.ClientRectangle.Width - 4, this.ClientRectangle.Height - 4);
```
ilexp: This is more of a style thing, but can you check how it looks when increasing padding on top and bottom of the editor? Maybe something between 5 and 20 pixels both top and bottom. Ignore if it doesn't turn out to look better.

Source/Plugins/EditorBase/PropertyEditors/GameObjectOverviewPropertyEditor.cs
```diff
@@ -120,6 +122,26 @@ protected void UpdateComponentEditors(GameObject[] values)
...
+			{
+				this.addComponentEditor = new AddComponentPropertyEditor();
+				this.addComponentEditor.Getter = this.GetValue;
+				this.addComponentEditor.Setter = this.SetValue;
+				this.AddPropertyEditor(this.addComponentEditor, this.ChildEditors.Count);
+			}
+			else
+			{
+				this.RemovePropertyEditor(this.addComponentEditor);
+				this.AddPropertyEditor(this.addComponentEditor, this.ChildEditors.Count);
```
ilexp: Why are you doing the add-remove here? Is it to ensure the editor remains at the bottom as new editors are added? If that's the case, please add a comment above those two lines, so it's clear to the reader.

Source/Plugins/EditorBase/PropertyEditors/AddComponentPropertyEditor.cs
```diff
@@ -0,0 +1,113 @@
...
+
+			this.ControlRenderer.DrawButton(
+				e.Graphics, this.rectButton,
+				buttonState, "Add Component");
+		}
+		protected override void OnMouseMove(MouseEventArgs e)
+		{
+			base.OnMouseMove(e);
+
+			bool buttonContainsMouse = this.rectButton.Contains(e.Location);
```
ilexp: We don't have this as a use case right now, but to future-proof this, we should already account for `this.ReadOnly`. Readonly should:

- Not react to mouse hover states.
- Not react to keyboard input.
- Display the button in deactivated state.

Source/Plugins/EditorBase/PropertyEditors/GameObjectPropertyEditor.cs
```diff
@@ -366,6 +391,34 @@ private void OnPrefabLinkBreakPressed()
...
+
+				this.PerformGetValue();
+				this.ParentGrid.Invalidate();
+			}
+
+			// Keeps add component button from being locked to the pressed 
+			// state caused by the SelectionDialog being modal
+			this.curButton = -1;
+			this.curButtonHovered = false;
+			this.curButtonPressed = false;
```
ilexp: 👍 for the explanatory comment and dealing with that case.

Source/Plugins/EditorModules/SceneView/Modules/SceneView.cs
```diff
@@ -678,9 +678,16 @@ protected void InitContextMenu()
 							ActionHandler	= this.gameObjectToolStripMenuItem_Click
 						},
 						new MenuModelItem
+						{
+							Name            = typeof(Component).Name,
```
ilexp: Can we add a "..." here to indicate that selecting this menu item will prompt a dialog for user input, rather than immediately performing an action? Would clarify what this item does a bit more by sticking to the "Windows convention".

deanljohnson:
I've addressed your points except the name of the dialog:

ObjectRefSelectionDialog: doesn't work because it is not generic enough and doesn't always select references
ObjectSelectionDialog: my issue with this one is that it does not necessarily select objects (unless we are talking objects as in .NET objects, but that is just too generic).
SearchableListSelectionDialog: I feel like the fact that it is searchable is a detail not need in the name
SelectionDialog: too generic as to the type of selection taking place

How about ListSelectionDialog?

ilexp:
Good points for all the names. Let's go with ListSelectionDialog for now.

ilexp:
Will check out the new version 👍

ilexp:
Merged. Thanks! 👍

Issue #647, opened 2018-05-01
SceneView "New / GameObject" item does not respect multi-selection


deanljohnson:
### Summary

Right clicking on the SceneView panel and selecting "New / GameObject" only creates an child object for the object clicked on when multiple objects are selected. I'd expect all objects selected to receive a new child object.

### How to reproduce [Bugs]

Select multiple objects within the SceneView panel. Then right click on one of them and select "New / GameObject". Only the object you right clicked on will receive a child.

ilexp:
Yep, that sounds like something we should change. Would expect the same behavior, but apparently it was never implemented that way.

Is multi-selection respected when creating new components for existing objects?

PR #649, opened 2018-05-02
647 sceneview multi object create

deanljohnson:
### Summary
Addressed #647 so that ""New/GameObject" now creates new child objects for every object selected.

There exists a minor issue: when multiple objects are created, only the last one created will have its name selected and in the process of being edited. I don't think any functionality exists to have multiple object names edited at the same time. What might make more sense is to not select a name for editing when multiple objects are selected. The other option would be some kind of popup dialog that asks for the name of the new objects.

### DIFF
--- a/Source/Plugins/EditorModules/SceneView/Modules/SceneView.cs
+++ b/Source/Plugins/EditorModules/SceneView/Modules/SceneView.cs
@@ -1477,16 +1477,22 @@ namespace Duality.Editor.Plugins.SceneView
 
 		private void gameObjectToolStripMenuItem_Click(object sender, EventArgs e)
 		{
-			// Create the GameObject
-			GameObject obj = this.CreateGameObject(this.objectView.SelectedNode);
-			GameObjectNode objNode = this.FindNode(obj);
+			// Create new objects
+			List<GameObjectNode> newObjects = new List<GameObjectNode>();
+			foreach (TreeNodeAdv node in this.objectView.SelectedNodes)
+			{
+				GameObject obj = this.CreateGameObject(node);
+				GameObjectNode objNode = this.FindNode(obj);
+				newObjects.Add(objNode);
+			}
 
-			// Deselect previous
 			this.objectView.ClearSelection();
 
-			// Select new node
-			if (objNode != null)
+			// Select newly created objects
+			foreach (GameObjectNode objNode in newObjects)
 			{
+				if (objNode == null) continue;
+
 				TreeNodeAdv dragObjViewNode;
 				dragObjViewNode = this.objectView.FindNode(this.objectModel.GetPath(objNode));
 				if (dragObjViewNode != null)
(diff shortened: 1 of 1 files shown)


ilexp:
Thanks! Putting this out for review.

> There exists a minor issue: when multiple objects are created, only the last one created will have its name selected and in the process of being edited. I don't think any functionality exists to have multiple object names edited at the same time.

Yep, that feature doesn't exist yet. I'd say it's fine for this PR - we can always add this later, and this is already an improvement compared to before.

> What might make more sense is to not select a name for editing when multiple objects are selected. The other option would be some kind of popup dialog that asks for the name of the new objects.

Both good suggestions. Another option would be to set a special flag for this specific operation, wait for finishing the rename and then apply the new name to all previously created objects.

Source/Plugins/EditorModules/SceneView/Modules/SceneView.cs
```diff
@@ -1477,16 +1477,31 @@ private void lockedToolStripMenuItem_Click(object sender, EventArgs e)
 
 		private void gameObjectToolStripMenuItem_Click(object sender, EventArgs e)
 		{
-			// Create the GameObject
-			GameObject obj = this.CreateGameObject(this.objectView.SelectedNode);
-			GameObjectNode objNode = this.FindNode(obj);
+			// Create new objects
+			List<GameObjectNode> newObjects = new List<GameObjectNode>();
+			if (this.objectView.SelectedNodes.Count <= 1)
```
ilexp: Could this block be unified with its else block? At a glance without actually stepping through code, it looks like the else branch covers the general "N objects" case, including the "1 object" case of the if block. Can you check if a unified variant works as expected?

Source/Plugins/EditorModules/SceneView/Modules/SceneView.cs
```diff
@@ -1477,16 +1477,31 @@ private void lockedToolStripMenuItem_Click(object sender, EventArgs e)
...
+				}
+			}
 
-			// Deselect previous
 			this.objectView.ClearSelection();
 
-			// Select new node
-			if (objNode != null)
+			// Select newly created objects
+			foreach (GameObjectNode objNode in newObjects)
```
ilexp: The new loop ensures all view items matching object nodes will be selected, which is exactly as expected. However, it also calls EnsureVisible and BeginEdit on all of them, even though only the last one will have a lasting effect. Can you adjust the code, so only the last (or first) view node gets those two calls?

deanljohnson:
Both good points - should be fixed now. That redundant if/else is a funny one - I distinctly remember writing that and thinking it was such a shame that I couldn't combine the if/else blocks. Don't know why I thought I couldn't lol

ilexp:
Damn, you were right about not being able to unify the if and else blocks - just found during testing that it now no longer supports the case when no object was selected. My bad! Will fix this on master.

ilexp:
Fixed. Binary release in progress. Should be out there in about 15 minutes.

Issue #650, opened 2018-05-06
Sprite Atlas Slicing Improvement


deanljohnson:
I've been working on an editor plugin called [Pixmap Slicer](https://github.com/deanljohnson/duality-pixmap-slicer) that can auto-slice a sprite sheet and allows the definition of rectangles with the mouse rather than by manually entering coordinates. 

After a discussion with @ilexp it was decided that we should integrate this into Duality directly. Additionally, it was noted that changes coming with v3.0 affect the integration of this plugin.

### Features (current and planned)
- Auto-slice sprite sheets
- Add/delete/edit rectangles with mouse
- Specify the order of rectangles in the atlas
- Slice the Pixmap with a grid from the slicing window
- Zoom and pan the Pixmap
- Naming of animation sequences or individual sprites

All of these features can be implemented without changing anything in the rest of Duality except for the naming of animation sequences.

### Integration Plan
1. [x] Cleanup current features and get a working baseline.
2. [x] Move changes into a forked Duality branch. PR to Duality master.
3. [x] Implement features onto Duality master that do not require atlas changes
4. [x] Move the slicer plugin to v3.0 branch and integrate with existing 3.0 changes
5. [ ] Improve atlas definitions to add support for named sequences, see issue #697 and PR #708
6. [ ] Implement additional features on the slicer given the named sequence support

deanljohnson:
### Progress
- Implemented add, delete, modify of rects
- Finished auto-slicing
- Improved zoom
- Added the ability to define the order of rectangles by clicking on them in the desired order
- Made the whole design much more extensible
- Documented current baseline and general cleanup

I'm going to start working on moving this to a forked Duality branch now. I want to get the initial PR in sometime soon so I can get some feedback on it.

@ilexp You had said this should be placed in the EditorBase Duality project. I'm looking at the folder structure there and I'm not sure where many of these things should go and don't want to start creating a bunch of new folders in that project. Here is a summary of the files I currently have and where they could go:

1x editor action -> EditorActions folder
The slicing form itself -> a new Forms folder
4x states (different modes of interaction with the slicing form, kind of like CamViewState) -> not sure where these should go. Maybe a Forms/PixmapSlicer folder?
1x event args file -> either new EventArgs folder or Forms/PixmapSlicer
2x utility files -> not sure

ilexp:
## Progress

- Concluded work on the v2.X PR for the Pixmap Slicer, merged back to `master`.
[...]

PR #651, opened 2018-05-07
Pixmap Slicer

deanljohnson:
### Summary

Integrated the [Pixmap Slicer](https://github.com/deanljohnson/duality-pixmap-slicer) plugin into the EditorBase Duality project as discussed in #650. I am submitting this PR to get feedback on the current state - I am sure there are quite a few changes to make before it is ready to merge.

### Features
- Auto-Slicing of a Pixmap with configurable alpha cutoff
- Add/delete/modify rects
- Zoom/Pan the Pixmap during slicing
- Rects can be ordered by clicking on them in the desired order

You can access the slicing window by right-clicking any Pixmap resource and selecting "Slice Pixmap". To modify a rect, click on it and then drag its edges around.

### Known Issues
- Closing the editor with a slicing form active and re-opening it will leave you with a useless slicing form. Selecting "Slice Pixmap" should always use an existing slicing form if possible.
- Lack of IHelpProvider implementation for advisor view
- Tooltips need to be improved
- Dragging the bottom/right past the top/left of a rectangle doesn't work right
- Colors/Look-and-Feel of slicing form don't quite match the rest of Duality

### Screenshots
Default View (image not included):
image

After Auto-Slicing:
image

Adding a New Rect:
image

Defining Rect Order:
image


deanljohnson:
### Progress
- Added Undo/Redo support
- Added IHelpProvider implementations
- Fixed negative width/height bug

### ToDo
- Tooltips
- Reuse existing Pixmap slicing form when possible
- Make sure look-and-feel matches other Duality windows as close as possible
- Form/EditorAction Icons
- Make sure all string are in resources where possible
- Extract pixmap slicing algorithm somewhere reusable

deanljohnson:
### Progress
- Pixmap slicing algorithm extracted from the form subfolder
- Tooltips for non-obvious buttons
- Icon for the window and editor action popup
- Moved many displayed strings into resources. Not finished here.
- Pixmap slicer forms are now re-used

### ToDo
- Finish moving displayed strings to resources
- Make sure look-and-feel matches other Duality windows

ilexp:
Good progress and writeup 👍 Give me a few days and I'll look into it to provide feedback.

ilexp:
> The "delete all" and "order rects" (the one with the hash) icons I created in GIMP by modifying one of the Silk icons. I'm not quite liking the "order rects" one yet.

The icons look somewhat blurry - are they scaled by the toolstrip somehow?

ilexp:
(You still seem to be working on this - ping me when you'd like to wrap stuff up for a review and merge)

deanljohnson:
Actually, I'd say it's ready for a review now.

deanljohnson:
So... I've never really taken a look at Duality's Tileset Editor. After doing so, I've determined some new features I'd like to add to this Pixmap Slicer to match the Tileset Editor in look and feel and to just enhance it's capabilities in general:

- Toggle Background
- Show Tile Indices
  - None, hovered, and all, just like Tileset Editor
  - Also, make styling of Pixmap Slicers indices match that of the Tileset Editor
- Buttons for zoom in/out/reset
- Apply/Revert
  - This is more minor as I feel undo/redo support is probably enough for the Pixmap Slicer (which it already has)

I'm going to get to work on those now :)

deanljohnson:
image

Done. Mirrors the Tileset Editor quite closely now. I copied over the icons that the Tileset Editor uses into the EditorBase project. I'm not sure if there is somewhere common they should be moved instead. I didn't bother adding the apply/revert stuff - I feel like the edits inside of the Pixmap Slicer are just a lot simpler than those in the Tileset Editor and so Undo/Redo is enough.

ilexp:
> [image]

Wow. Really looking forward to this. Nice work 👍 

> I didn't bother adding the apply/revert stuff - I feel like the edits inside of the Pixmap Slicer are just a lot simpler than those in the Tileset Editor and so Undo/Redo is enough. And, there is no resource compilation needed with Pixmap Slicer.

Agree completely, good call.

ilexp:
Gave it a quick first look, solid piece of work so far 👍 Here's a first list of change / feature / bugfix / polishing requests:

- Some namespaces are quite extensive and contain classes that will likely conflict with others due to somewhat generic names - can we use more specialized names and shorter namespaces instead?
  - Specifically looking at `Utilities.cs`, `Side.cs` and `PixmapSlicing.cs`.
  - Would prefer if we can avoid having a separate namespace for pixmap slicing.
  - Maybe rename `PixmapSlicing` to `PixmapSlicingUtility` and unify with `Utilities`, moving both of them into the `Duality.Editor.Plugins.Base` namespace.
  - Consider renaming `Side` to something more specific (like `PixmapSlicerRectSide`), or make it a private nested class of the slicer form, so it doesn't need its own namespace to exist.
- When changing states in the slicer form, the tool buttons visibly disappear and pop into existence. There's probably a spot where `SuspendLayout` / `ResumeLayout`, or its equivalent for toolbars needs to be called.
- Dragging a rect side feels somewhat slow, and I get an audibly high CPU (fan) load from doing that. Something may be off with performance here. Maybe it's not actually the drag operation, but the redraw of the control, not sure.
- Zooming in on a previously centered sprite sheet will spawn scroll bars, even if it would still fit comfortably into the view area.
- Hitting the "Show Indices Display Mode" button after zooming in and out will re-center the image, showing that it was off center before.
- The dark background mode needs other colors for the slice rects in regular / selected state, as they are hard to read "dark on dark".
- When a rect is selected, you can hover and drag its edge beyond its corners. To test this, switch to "show index on hover" mode and note how the index is displayed when hovering beyond the corner. Click + drag works there as well.
- The sprite sheet extents / border should be visibly displayed, so the user can distinguish between transparent areas and regions outside the sprite sheet.
- It's unclear to me as a user what happens when I start specifying the order of atlas rects, but then hit the "Ok" button before I specified all of them. I still get the full number, and their order did change, but I don't know how all others are ordered. Should we disallow this, or maybe switch on rect index display as an FYI?
- My "Show Indices Display Mode" is discarded after completing a manual order setup.
- The grid slicer should probably have a min setting of one row and one column. Wouldn't know what to use zero for, and it allows to enter an inconsistent state when reducing one setting to zero and back up after specifying the other. Also, it's possible to get a NullReferenceException when attempting to complete grid slicing with zero rows and zero columns.
- The image is displayed a bit blurry, even though no zoom is active and it fits entirely into the view - can we get this to work pixel-perfect in those cases, as well as all Nx (2x, 3x, 4x, ...) zoom levels?
- I can switch which Pixmap to edit in an already open PixmapSlicer by simply right-clicking the other and invoking the slice action. Can we simplify this and let the slicer window automatically react to editor selection? That way, I'd just need to select the other Pixmap and could begin slicing immediately.
- When zooming in, the rects [get a bit erratic in their width](https://i.imgur.com/YJZrtbD.png), which feels buggy. Can we get them to always be one pixel wide regardless of zoom?
- Also, I think it would be great if dragging a rect side would snap, so the resulting atlas rect is always defined in full pixels.
- It is currently possible to zoom with the mouse wheel, which is really nice. However, the zoom will not respect the current view center, so zooming in will always "move the view" towards the upper left, and vice versa. It would be great if the zoom was always towards the current view center.
- It is currently possible to drag the side of a rect until it's below zero size. The rect will disappear and only be accessible in the Object Inspector, or via "Delete All".

I really like the auto-zoom on resize, how the inspector updates in realtime, the full undo / redo support, how it all integrates with the rest of the editor, and of course auto and grid slicing itself. So there's some polishing left to do, but still really nice work overall 👍

deanljohnson:
> Some namespaces are quite extensive and contain classes that will likely conflict with others due to somewhat generic names - can we use more specialized names and shorter namespaces instead?

I improved the Side and Utilities namespaces. I kept the states in their own namespace as there are quite a few classes there and no one else should ever really need to reference them.

> When changing states in the slicer form, the tool buttons visibly disappear and pop into existence. There's probably a spot where SuspendLayout / ResumeLayout, or its equivalent for toolbars needs to be called.

Good catch. Fixed.

> Dragging a rect side feels somewhat slow, and I get an audibly high CPU (fan) load from doing that. Something may be off with performance here. Maybe it's not actually the drag operation, but the redraw of the control, not sure.

Pretty sure its the redraw resizing causes a lot of invalidation events. I tried at one point to invalidate only the necessary region rather than the whole form and it didn't make much difference. I'll take another look at this.

> When a rect is selected, you can hover and drag its edge beyond its corners. To test this, switch to "show index on hover" mode and note how the index is displayed when hovering beyond the corner. Click + drag works there as well.

Can't believe I missed that bug for so long... it was an error in how I was calculating distance to a rect from a given mouse position. Fixed

> The sprite sheet extents / border should be visibly displayed, so the user can distinguish between transparent areas and regions outside the sprite sheet.

Only the image background uses the checkboard now. The rest of the display will be a gray color.

> It's unclear to me as a user what happens when I start specifying the order of atlas rects, but then hit the "Ok" button before I specified all of them. I still get the full number, and their order did change, but I don't know how all others are ordered. Should we disallow this, or maybe switch on rect index display as an FYI?

I definitely don't think it should be disallowed. The advisor window provides a help message here saying that all other rects will be placed at the end of the atlas. It maybe should say that it is in no particular order. Technically there is some order - it is a stable reordering so the relative order of unclicked rects will be maintained afterwards.

> The image is displayed a bit blurry, even though no zoom is active and it fits entirely into the view - can we get this to work pixel-perfect in those cases, as well as all Nx (2x, 3x, 4x, ...) zoom levels?

Hmm, the graphics object is being told to display the image at the highest detail possible so I'm not sure why it would look blurry in that case. When zoomed there is going to be some blur applied of course, but when unzoomed there should not be. I haven't noticed any issue myself but I'll see what I can find out about how images are drawn by `Graphics`.

> I can switch which Pixmap to edit in an already open PixmapSlicer by simply right-clicking the other and invoking the slice action. Can we simplify this and let the slicer window automatically react to editor selection? That way, I'd just need to select the other Pixmap and could begin slicing immediately.

Done.

> Also, I think it would be great if dragging a rect side would snap, so the resulting atlas rect is always defined in full pixels.

Done.

> It is currently possible to drag the side of a rect until it's below zero size. The rect will disappear and only be accessible in the Object Inspector, or via "Delete All".

Fixed. Can't make rects with zero width of height anymore.

> All comments about mouse wheel scroll, zooming, and scroll bars.

I have had a terrible time getting these things to work. I tried to address most of your comments and one point or another and always ran into trouble. As changes to everything else is starting to settle down I will take another look.

deanljohnson:
I think I have addressed all of your points except for the high CPU load when dragging. In order to make it seem responsive, it needs to invalidate the control with every bit of movement. I tried invalidating only the area that changed and it made no difference while adding quite a bit of mess to the code.

You also said that the resizing feels slow along with the comment about high CPU load. That seems to have been fixed when I made it snap to pixel coordinates.

Regarding the image quality - it is far better now but it still has a very minor blur to it - pretty hard to notice. To achieve this quality, I had to make sure the image preview was generated at the displayed resolution, not at a lower resolution and then scaled up as it was before. I also set the interpolation mode to `NearestNeighbor` which should be giving the minimum amount of blur. I think the current blur is coming from `PreviewProvider` as it seems to use linear interpolation when scaling the image.

ilexp:
Wow, you're really fast. Those were quite a few points! 

> I think the current blur is coming from PreviewProvider as it seems to use linear interpolation when scaling the image.

Honestly I wouldn't use `PreviewProvider` here at all, as those previews aren't intended for generating the primary view of any editor. They're more of a "give me a quick visual impression" API. If you're doing a specialized editor, prefer a specialized way to draw your view instead.

> Note: Anything I didn't respond to above has been fixed, I just simply didn't have anything to add :)

Nice work 👍 

> I think I have addressed all of your points except for the high CPU load when dragging.
> ...
> Regarding the image quality - it is far better now but it still has a very minor blur to it - pretty hard to notice.

Will get back to you as soon as I manage, and maybe take a look into the performance and blur issue myself to help out where I can.

deanljohnson:
I can't really remember why I started with using PreviewProvider... well its gone now and so is the blur!

I also did two things to improve resizing performance - resizing only invalidates the updated region and rects are now drawn as a batch. Unfortunately it didn't make a huge difference. I'm seeing about a 10% improvement. After running it through a profiler I can see that the execution time is almost entirely spent within the various `Graphics` methods. Nothing else really shows up on the profile.

ilexp:
> All in all, with this fix and the above performance improvements, dragging is at least 40% faster.

Nice work 👍 Didn't find the time to re-review yet, but hopefully will soon.

deanljohnson:
> Next, I'd like to distinguish between Nx (2x, 3x, ...) ratio zoom levels and other zoom levels:

Handling the fractional and Nx zoom levels differently seems very strange to me. I'd prefer to remove fractional zoom support and just have Nx zoom levels. That would be simpler and still good from a usability perspective - we aren't writing an image editor so we don't need fine grained zoom control if it is going to be an issue.

ilexp:
> I'd prefer to remove fractional zoom support and just have Nx zoom levels. That would be simpler and still good from a usability perspective - we aren't writing an image editor so we don't need fine grained zoom control if it is going to be an issue.

I'd be fine with that too. We still need 1/Nx zoom levels though, so it's possible to see the full image and get a quick overview on what the auto-slicing has done, and those need to be properly filtered when scaled down, so we won't dodge that filtering switch entirely.

ilexp:
Worked on this PR a bit more and actually started with a few bigger changes. Still work-in-progress:

When I started addressing the zoom issue, I noticed that all the drawing, scrolling and zooming was handled directly by the docking control, which looked a bit heavy-handed and might fall on our feet later on: Drawing would need to know the layout of the form in order to draw in the right places, scrolling was done manually and invalidation logic would have to be hand-optimized for scrolling scenarios. Also not the typical design you'd expect WinForms wise.

To address this, I gathered all UI-related functionality and created a new `PixmapSlicingView` WinForms control to house them. This control is intended to work stand-alone, while the form and its states uses the views public API to control it, and add editing functionality on top. This is in line with other places in the editor (such as the Tilemaps editor plugin), and also makes maintaining and re-using UI code a bit easier.

With the new WinForms control, a lot of state-related code could be simplified by simply allowing the editing state to access the view control and its API in order to change settings or transform coordinates. Changing zoom levels no longer re-generates the background image and scrolling re-uses 90% of the previously drawn image and only invalidates the regions that just entered view. It is also handled by WinForms for the most part.

There are, however, a few regressions due to the fact that I haven't had as much time for this as I'd like. I've updated the ToDo list to reflect this and will get back to this as soon as I manage. 

## ToDo

- Regression: The image is no longer centered. Fix this.
- Regression: Index labels flicker when dragging rect edges while zoomed in, due to hover state not always matching selection state. Address this by moving selection and edge hovering logic to the `PixmapSlicingView` and letting the editing state use and configure it.
- Regression? Grid slicing now breaks the existing atlas when cancelled.
- Zooming in should "snap" to Nx zoom levels.
- Re-review latest state and merge back when approved.
- Merge from `master` to `v3.0` branch.

If you have any thoughts on this, they're of course appreciated.

deanljohnson:
I was planning on doing essentially the same refactoring you did here!

To address the regression of the index labels flickering we would just need to force set the hovered rect to the selected rect whenever a selected rect edge is hovered. That's how I had addressed it before and it worked well and was simple.

> Regression? Dragging the left or top edge back and forth while zoomed in very far will move the right / bottom edge and shrink the rect.

Sounds like a rounding issue.

I have been short on time myself the last week and probably won't have time over the next few days. When I get the chance I'll take a look and see if I can address any of the ToDo items.

ilexp:
> I was planning on doing essentially the same refactoring you did here!

Great to hear 🙂 I managed to put some more work into this. Things start to take shape again.

> Sounds like a rounding issue.

That was exactly it. I moved rect clamping and adjustments from view space (which was partially int-locked) into atlas space and the problem no longer appears.

> To address the regression of the index labels flickering we would just need to force set the hovered rect to the selected rect whenever a selected rect edge is hovered. That's how I had addressed it before and it worked well and was simple.

I addressed this by moving all of the hover and selection logic into the `PixmapSlicingView` control, which knows best about that anyway. The editing states now use this information to do their thing, but no longer have to figure out selection and hovering on their own. 
With the view now having all of the knowledge in one place again, the fix basically emerged by itself. Very similar, but instead of adjusting the hovering info to match selection, it now displays the indices of selected rects as if they were hovered. That way, you can still get the correct hover info when needed, regardless of selection

----

Other than the above, there are a few minor API and style changes as usual, and one bigger changes in how Undo / Redo is addressed:

The way I've seen this in the code before, the Pixmap was edited as needed during operations, and then, when finished, committed. This commit involved rolling back to its original state and adding a final UndoRedo action. Makes a lot of sense for dragdrop, as this is one continuous operation you don't want to turn into a stream of UndoRedoActions. However, this has the disadvantage of producing an inconsistent state, where the resource was already edited, but no action was filed yet - and it also requires an explicit "commit" in the end, along with a guarantee to run into it in all cases. 

There is a different way to tackle this in Duality, which avoids the downsides of it: Append-able UndoRedoActions. Basic idea is, when filing a new action, it first checks whether it could be appended to the previous registered one. If this is the case, it will be appended / merged into it without producing a new undo stack entry. That way, continuous operations can just append a new action on every step, and the system handles and registers them all as a single action. When done, an optional call to `UndoRedoManager.Finish` will avoid further merges into the previous action to allow for two distinct dragdrop operations in sequence. [Here's my changes](https://github.com/AdamsLair/duality/pull/651/commits/413d44a8b559d065127f9230696e81ad88e57a64#diff-e2c3bca2f072dd28cddcecbbae2172daR27) to the `SetAtlasRect` action as an example on how to make an UndoRedoAction append-able.

## ToDo

- Regression: The image is no longer centered and no longer retains its center-based "viewport position". Fix this.
- Regression? Grid slicing now breaks the existing atlas when cancelled.
- Zooming in should "snap" to Nx zoom levels.
- Re-review latest state and merge back when approved.
- Merge from `master` to `v3.0` branch.

> I have been short on time myself the last week and probably won't have time over the next few days. When I get the chance I'll take a look and see if I can address any of the ToDo items.

Same here, so no worries, I know how it is. Feel free to pick this issue up again at any point when you have time. I'll do the same and we'll be done when we're done.

deanljohnson:
@ilexp I've found a few minor bugs that I'm addressing, but I found a different issue that I think warrants discussion.

Originally, you had to select a rect before you could resize it. Now, as long as your mouse is hovering an edge you can start a resize operation by clicking and dragging. When dealing with two rects with overlapping edges this produces a problem as you can have one rect selected and then when trying to click and resize that rect the code determines the other rect is the one you want to use and starts resizing that. I think there are two ways of solving this:

1. Revert to old behaviour. Only edges of the selected rect can be hovered and thus resized.
2. Maintain current behaviour where any edge can be hovered, but prefer using the selected rect when multiple edges are hovered.

I'd prefer the first option because it is more consistent from a usability perspective. It is also simpler to implement.

Thoughts?

ilexp:
Short on time, so I'll try to be brief 🙂 

> 1. Revert to old behaviour. Only edges of the selected rect can be hovered and thus resized.
> 2. Maintain current behaviour where any edge can be hovered, but prefer using the selected rect when multiple edges are hovered.
>
> ...
> 
> Thoughts?

I'd prefer to go with option 2, as requiring to select first is an additional action / click that, from my personal user perspective, feels unnecessary. In most cases, I already select the edge I want to drag in my first click and would expect to be able to immediately move it without an additional interaction first.

> Edits: Bugs I had mentioned are fixed.

Nice! 👍

deanljohnson:
> I'd prefer to go with option 2, as requiring to select first is an additional action / click that, from my personal user perspective, feels unnecessary. In most cases, I already select the edge I want to drag in my first click and would expect to be able to immediately move it without an additional interaction first.

Alright I'm fine with that. While looking into this I ran into another issue. When your mouse pointer is within two rects, only the highest indexed one will be hoverable even when the mouse pointer is clearly over a side of the lower indexed rect. This is because hovering over the center of a higher index rect is taking precedence over the side of a lower indexed rect because it break out of the search loop. I'm looking into how to cleanly address this. Ideally, hovered a rect side should always take precedence I think.

ilexp:
Sounds good! 

Will get back to the issue as soon as I manage. Let's see if we can start to wrap things up.

ilexp:
## Progress

- Merged this PR to master.
- Triggered a binary release.
- Merged master to develop-3.0.

Issue #16, opened 2013-07-21
Copy / Paste for GameObjects and Components


AdamsLair:
- Instead of the current Clone functionality, implement a proper Cut / Copy / Paste
- Use the Clipboard
- Apply to SceneView and CamView
- Use Case: Copy GameObject and Paste it somewhere else (Hierarchially)
- Use Case: Copy Component and Paste it into a different GameObject
- ...

ilexp:
Done by @deanljohnson. Closing this.

PR #652, opened 2018-05-10
Copy/Paste GameObjects and Components in SceneView

deanljohnson:
### Summary
Implemented copy and paste functionality for GameObjects and Components within the SceneView (see #16). In the process, removed cloning of GameObjects from the SceneView context menu.

### Details
- Added new PasteGameObjectAction in DualityEditor
- Added new context menu actions to SceneView for Copy and Paste
- Removed clone context menu action
- Supports copy/paste of arbitrary object hierarchies and can paste onto multiple objects

While modifying the project Resources, VS regenerated the SceneViewRes.Designer.cs file and switched from German(I think) to English. I wish it wasn't dirtying up this PR with unrelated changes but I don't know how to fix it but keep the changes it needs... Sorry.


ilexp:
Nice work, glad to see the old issues tackled as well 👍 

> While modifying the project Resources, VS regenerated the SceneViewRes.Designer.cs file and switched from German(I think) to English.

Yep, that was German - but no worries. Switching this to English is a good thing, might as well do it as a side effect on this PR, since it's an automated conversion anyway.

However, there's a design flaw in the issue we need to figure out first: As far as I can see, you're putting the copied objects references in the clipboard. That means: 

- I can copy an object, delete it, and then paste it, which would clone a disposed object back into existence, either failing or producing one in an invalid state. 
- I can also copy an object, modify it, and paste it, which would yield a new object in the modified version, which is not the one I copied.

For proper copy-paste support, we'll need to serialize the copied objects for the clipboard, and deserialize them afterwards. Serialization alone could still be a problem though, since it will pull any referenced objects into the process, essentially causing the entire scene to end up in the clipboard in a worst case scenario. Cloning is a few steps ahead here, since it actually figures our an ownership graph, but cloning also is a live process. We need a combination of both here.

I think the first approach I'd try here would be to see if it is possible to do a clone operation on copy, but only in order to create an object graph that has no external references. Cloning itself, unfortunately, re-establishes external references instead of nulling them, but there may be mechanisms that could be used to step in here and prevent that. The resulting cloned objects would then be ready to serialize without the danger of accidentally pulling other objects into it.

Now, regarding the serialization part, I'd say we could just use the XmlSerializer, and put the resulting Xml in both a special clipboard channel, and the text one, so we'd get the gimmick of being able to copy a GameObject and paste it into a text editor, as well as copy an XML-serialized object and paste it into the scene. We'd also gain the ability to copy-paste between different Duality instances, to a degree even if they use different Duality core versions.

I realize the copy / paste mechanism I outlined is probably bigger than anticipated, just let me know whether you're interested in pursuing this further. Also, would be interested in your thoughts about this outline.

deanljohnson:
> As far as I can see, you're putting the copied objects references in the clipboard.

I don't think that's the case. I am specifically making sure to call DeepClone on all objects/components as they are placed on the Clipboard specifically to address the two bullet points you listed. Is that not solving those two points?

Regarding re-establishing external references - I think in most cases we want those references re-established as they are part of the copied GameObject/Components data. If those references have been deleted, the deletion of them would have pushed a UndoRedo action onto the stack so the copy/paste operation can't be undone until the deletion is undone.

I'm not totally familiar with how the cloning stuff works in Duality, but I would guess that it nulls/invalidates fields or references it is unable to establish which is what we want.

I don't know if I understand your points properly. Can you give me a case that I could try out to expose a problem in the code? In really complex hierarchies I could imagine there being a case where this fails, but so would cloning and so I don't see how it's really any different than what we had before.

Source/Plugins/EditorModules/SceneView/Modules/SceneView.cs
```diff
@@ -365,37 +367,94 @@ protected Node ScanScene(Scene scene)
...
-				select (objNode as GameObjectNode).Obj;
-			var objArray = objQuery.ToArray();
-			
+			var nodeList = nodes
+				.Select(n => this.objectModel.FindNode(this.objectView.GetPath(n)) as NodeBase)
+				.ToList();
+
+			var objects = nodeList
+				.OfType<GameObjectNode>()
+				.Select(gon => gon.Obj.DeepClone());
```
deanljohnson: Here is the GameObject cloning and a few lines below is the Component cloning.

ilexp:
> I don't think that's the case. I am specifically making sure to call DeepClone on all objects/components as they are placed on the Clipboard specifically to address the two bullet points you listed. Is that not solving those two points?

Ohh.. got it. Didn't see the copy and was approaching the issue from a different angle that made solving this more difficult. For some reason I was under the impression that you store the reference and do the clone only when _pasting_, not when copying. In fact, doing the clone first and then referencing the cloned instance should solve the two points just fine. Thanks for pointing it out 👍 

I'll put this out for community review, or take a look myself soon-ish, whoever gets there first 🙂

deanljohnson:
Sounds good - I was originally placing references on the Clipboard but had the same issues you brought up. I was then going to serialize to XML and place in a DataObject, but found the extension methods you wrote to place an IEnumerable<GameObject/Component> into a DataObject and used that.  Hopefully that keeps things simple. I can't think of a way this would fail where cloning wouldn't (since all this really does is clone....) so this should work at least as well as the previous cloning stuff.

deanljohnson:
There is more work to be done with this one. If you have this hierarchy:

```
-GameObject A
  -GameObject B
    -Component with a reference GameObject A
```

And you then copy and paste GameObject B somewhere, the reference will be to a non-existent object. I think either nulling the reference or referencing the actual GameObject A are both valid implementations, but the current state is just broken. This is all related to #653 so until that is fixed this will not be mergeable.

ilexp:
> And you then copy and paste GameObject B somewhere, the reference will be to a non-existent object.

By non-existent, you mean an object is references which is not part of the scene graph?

> What I'm not sure about is whether or not this is the fault of the cloning procedure or the fault of serialization that occurs when using the clipboard as discussed in #653. I'm guessing the latter since you said earlier that cloning should establish external references. 

Assuming we're experiencing an "object outside scene graph" result here, I'd guess the fault is that serialization of the object graph occurs in the clipboard. You can verify this with a breakpoint in [SerializableWrapper](https://github.com/AdamsLair/duality/blob/012a34282f660f06292590e9b1addb6f8735cab9/Source/Editor/DualityEditor/Utility/SerializableWrapper.cs), which would be consistent with #653.

> To fix this we would either have to walk the objects/components being copy/pasted and null their external references or put all of them into an ID table such as discussed in #653.

I think the proper fix for this should be done as part of #653, as this could be the main culprit. If reference serialization through the clipboard was working / implemented, I would assume that the problem within this PR would vanish as well.

deanljohnson:
I've merged master (with the fix for #653) and it has fixed the bug I had mentioned.

deanljohnson:
> To avoid situations where copy / paste is wrongfully disabled, the context menu needs to be - at least partially - updated on selection changes as well.

I don't think I understand what you are saying here. The selection can't be changed while the context menu is open so I don't see what it would need to respond to selection changes. Are you saying that a context menu item with a shortcut can be disabled upon opening the menu, then a selection change can occur that makes the item valid, but the shortcut won't be usable until the context menu is opened again? If so, I just tried to cause that to occur and everything seemed to work. Do you have a specific test case to bring this out?

> I think we should keep the cloning option, maybe as Duplicate via Ctrl + D?

Sounds good. I'll work on it.

> *Clipboard stuff*

This is kind of sounding like a use case for storing multiple formats of data in the DataObject. Anything that puts an object/component onto the clipboard should store the item in a "cloned" format and a "reference" format. The paste location could then have logic to grab the format that it needs.

> It actually breaks right now when copying a GameObject or Component reference in the Object Inspector and pasting it in the Scene View: What was copied as just a reference to an existing object is pasted assuming that it is a detached clone.

How does it break? I just tried it now and it seemed to work fine.

deanljohnson:
> It actually breaks right now when copying a GameObject or Component reference in the Object Inspector and pasting it in the Scene View: What was copied as just a reference to an existing object is pasted assuming that it is a detached clone.

Ahh - I think you are talking the other way around - copying from the SceneView and pasting into a field in the object inspector. That doesn't work because of the cloning. In this case, I think having the multiple formats in the DataObject would fix this quite well as we could have the SceneView place both a reference and a clone into the DataObject and have the inspector only take the reference.

ilexp:
> I don't think I understand what you are saying here. The selection can't be changed while the context menu is open so I don't see what it would need to respond to selection changes. Are you saying that a context menu item with a shortcut can be disabled upon opening the menu, then a selection change can occur that makes the item valid, but the shortcut won't be usable until the context menu is opened again? If so, I just tried to cause that to occur and everything seemed to work. Do you have a specific test case to bring this out?

No no, what I mean is: Right-click without a selection to bring up the context menu without the copy item, then select a node and hit `Ctrl + C` - that should theoretically reproduce the problem. Since the context menu wasn't updated (as the menu wasn't opened in the meantime), the copy command is still inactive, so the keyboard shortcut is ignored.

> How does it break? I just tried it now and it seemed to work fine.

I got an exception somewhere in the tree view, but didn't look into it to be honest, since I already expected this case to fail. The case I tested was a nested RigidBody hierarchy, where the innermost body had an AngleJoint referencing the second-to-innermost body. I copied that one and pasted it into a different root object in the scene. 

It's a bit convoluted, but I think it should have the potential to break in any case where you copy a GameObject or Component reference from the Inspector and paste it in the Scene View - maybe there are some cases where it recovers though, like when the referenced object is a root object, which is simply re-inserted as a "new" root object.

> This is kind of sounding like a use case for storing multiple formats of data in the DataObject. Anything that puts an object/component onto the clipboard should store the item in a "cloned" format and a "reference" format. The paste location could then have logic to grab the format that it needs.

Sounds good. We'd need to find a way to allow cases where the copy site stores a reference and the paste site wants a clone - could always generate that one internally, if there's a reference available, or do it manually at the paste site. Anyway, let's not do this as part of this PR, could be big enough to be its own thing, along with some related cleanup.

ilexp:
> Ahh - I think you are talking the other way around - copying from the SceneView and pasting into a field in the object inspector. That doesn't work because of the cloning.

It actually breaks both ways, at least in some cases, see above - posted before I saw your new comment.

deanljohnson:
Put together a pretty simple implementation of my multiple formats idea. Allows copy/pasting between inspector and SceneView without issue. It's very basic but it would give us Copy/Paste for now. Could clean it up a tiny bit for a merge of this and then rework things on a larger scale in another PR?

ilexp:
Let's not rush this. The DataObject extensions code is a part of the general editor infrastructure - it's quite ugly in its current state, but I wouldn't want to make it worse with a quickfix for a specific UI use case, even temporarily. I'm fine to just leave this PR open until merging becomes viable, so the reference / serialized problem with the current DataObject extensions can be solved properly in time.

How we proceed here depends - if you're interested in redesigning that part of the DataObject handling and already have a good idea on how to do that, feel free to do a PR on this, and we'll iterate there. Otherwise, we could wrap this up in a new issue, to be discussed and started independently.

deanljohnson:
Sounds good. I'm writing up the issue now for the DataObject format stuff so we can discuss that implementation. If it's seems reasonable I can get started on the code side.

deanljohnson:
Will do.

deanljohnson:
Master has been merged.

I addressed the issue of available context items only being updated when the context menu was opened, rather than whenever the selection changes. I also had to make sure invisible context items were disabled because even when invisible they could still be invoked by shortcut.

There is a quirk to note in the difference between reference and value storage. When copying an object from the scene view, I had to store it as both a reference and as a value - reference for when copying the object and pasting into a reference field on a component and value for when pasting back to the scene view. I couldn't rely on the automatic ref->value conversion because we want the state of the object upon copying, not upon pasting, to be what gets pasted into the scene view. Nothing wrong so far. 

When we try to copy/paste an object reference (such as from a component property) into the scene view we can have an issue though. When copying from the object inspector it is only placing a reference into the DataObject. If the object being referenced is changed before being pasted, the pasted objects will have those changes. Fixing this would require that anything that places Components/GameObjects into a DataObject place them in both formats. Perhaps the DataObject extension methods for Components and GameObjects should do this automatically.

Thoughts?

ilexp:
> I addressed the issue of available context items only being updated when the context menu was opened, rather than whenever the selection changes. I also had to make sure invisible context items were disabled because even when invisible they could still be invoked by shortcut.

Ah - good catch!

> There is a quirk to note in the difference between reference and value storage. When copying an object from the scene view, I had to store it as both a reference and as a value - reference for when copying the object and pasting into a reference field on a component and value for when pasting back to the scene view. I couldn't rely on the automatic ref->value conversion because we want the state of the object upon copying, not upon pasting, to be what gets pasted into the scene view. Nothing wrong so far.

Okay, makes perfect sense.

> When we try to copy/paste an object reference (such as from a component property) into the scene view we can have an issue though. When copying from the object inspector it is only placing a reference into the DataObject. If the object being referenced is changed before being pasted, the pasted objects will have those changes. Fixing this would require that anything that places Components/GameObjects into a DataObject place them in both formats. Perhaps the DataObject extension methods for Components and GameObjects should do this automatically.

Hmm.. this is indeed a bit tricky, concept wise. Thinking about it, when the user copies an object reference from the inspector, pasting them in the scene view, where "the real thing" is expected, rather than "just a reference", this is somewhat of a concept breach. 

One way to handle this would be to disallow pasting references into the scene view, because it's "the wrong type" - not sure about that. Another way would be to always go both ways like you suggested. In fact, I think I'd favor the latter option and go both ways. Which begs the question: Should this be the default for wrapped objects?

The kind of problem we're encountering here could theoretically show up for any kind of object, not just GameObject and Components. Going both ways for wrapped data in general would mean changing the `DataObject` extensions slightly, so that setting it would always store both kinds of wrappers simultaneously, with the getter methods only deciding which one to look up. 

As usual, thoughts appreciated.

deanljohnson:
> One way to handle this would be to disallow pasting references into the scene view, because it's "the wrong type" - not sure about that. 

We would need to add back in the `allowConversion` flag to the DataObject methods to support this, but is certainly easy to do.

> Another way would be to always go both ways like you suggested. In fact, I think I'd favor the latter option and go both ways. Which begs the question: Should this be the default for wrapped objects?

I'm hesitant in making this the default for wrapped objects because I view that portion of the API as a thin addition to DataObject. If someone is calling those methods, they should have almost full control over what happens and what gets stored.

Currently, ContentRef, BatchInfo, and IColorData methods all use Value storage and do not have the option to store references. If we switch the wrapped data methods to always use both we need to be careful with these methods to make sure they continue working the same.

What may be better than always storing wrapped data in both forms is to store it in both only if a reference is given (do an immediate DeepClone). If someone passes `DataObjectStorage.Value` to `SetWrappedData` then we can assume that the caller does not want references maintained and we can only store as a value. I think that conceptually this makes sense - if a reference is placed in the DataObject and a value is later retrieved, I'd expect the value to be a clone of the object at the time of placing it into the DataObject.

In the case that we automatically store value forms of references, can we safely say that cloning is always safe to do even when that clone is potentially unused? I can't think of a case where there would be a problem but want to make sure.

ilexp:
> What may be better than always storing wrapped data in both forms is to store it in both only if a reference is given (do an immediate DeepClone). If someone passes DataObjectStorage.Value to SetWrappedData then we can assume that the caller does not want references maintained and we can only store as a value. I think that conceptually this makes sense - if a reference is placed in the DataObject and a value is later retrieved, I'd expect the value to be a clone of the object at the time of placing it into the DataObject.

Makes sense. I'm all for it 👍 

> In the case that we automatically store value forms of references, can we safely say that cloning is always safe to do even when that clone is potentially unused? I can't think of a case where there would be a problem but want to make sure.

Can't think of a case either. Even in cases where I (for whatever reason) clone something that has native resources allocated, I'd argue the cloning implementation and design of that specific object, or containing object, would be responsible for avoiding problems created by stray copies. I think it should be fine.

By the way, what happens on multi-select? If I select 1000 objects in the scene view and hit `Ctrl+C`, will I get a big `DataObject` containing 1000 references and 1000 serialized objects? We might need to keep an eye on performance here. As a user, I would expect a copy operation like that could take a perceivable amount of time - but we should check it's within reasonable margins, just to be sure.

deanljohnson:
Okay - so I implemented the automatic value storage of reference wrapped data.

Multi-select doesn't cause a problem. Copying a thousand objects is no problem and pasting them is fine unless you do something like the following.
1. Create a large number of game objects (100 should be plenty)
2. Select them all and CTRL + C
3. **With all of them still selected**, CTRL + V

The editor will be stuck for quite a while on this because it is trying to paste a clone of the operation into every single object - so with 100 selected you are then pasting 100 objects into each of them leading to 10,000 objects being pasted. The issue is not the number of objects or the copy/paste code but the node insertion code of the scene view. It performs each insertion individually and that is just too slow for an operation like this.

ilexp:
> It performs each insertion individually and that is just too slow for an operation like this.

Ah.. yeah, that makes sense. But that is a fix for a different issue / PR, let's wrap up this one 🙂 

Actually, I'm pleasantly surprised 10k objects is fine when it comes to the cloning operation alone. Would have expected worse.

> The editor will be stuck for quite a while on this because it is trying to paste a clone of the operation into every single object - so with 100 selected you are then pasting 100 objects into each of them leading to 10,000 objects being pasted.

That's actually a really neat feature. I don't think I've seen multi-select paste anywhere else yet. Nice work 👍 

I'll put this on my review and merge ToDo.

ilexp:
Fixed two minor bugs during testing and added the duplicate shortcut to Camera View as well, so you can continue to use the same shortcuts in both Scene View and Camera View. Code looks good, I think we're ready for merge 👍

Issue #653, opened 2018-05-10
Investigate and Potentially Fix Clipboard Handling of GameObject References


ilexp:
### Summary

While investigating PR #652, I looked through some old editor code that is used when calling `dataObject.SetGameObjectRefs(x)` and it seems that the current setup will clone and serialize the objects when storing them on the Clipboard, rather than storing just their references as documented.

### Analysis

- [SetGameObjectRefs](https://github.com/AdamsLair/duality/blob/012a34282f660f06292590e9b1addb6f8735cab9/Source/Editor/DualityEditor/Utility/Extensions/ExtMethodsDataObject.cs#L123) invokes [SetWrappedData](https://github.com/AdamsLair/duality/blob/012a34282f660f06292590e9b1addb6f8735cab9/Source/Editor/DualityEditor/Utility/Extensions/ExtMethodsDataObject.cs#L22), which packs the objects in a [SerializableWrapper](https://github.com/AdamsLair/duality/blob/012a34282f660f06292590e9b1addb6f8735cab9/Source/Editor/DualityEditor/Utility/SerializableWrapper.cs).
- As long as the `DataObject` is not put in the `Clipboard`, no serialization takes place and the objects are retrieved back as references, as expected.
- However, as soon as storing it on the `Clipboard`, serialization is performed, so new objects will be retrieved back, rather than references to the previous objects.
- Based on the method name and usage, expected behavior would be to always retrieve either references, or no value at all. There are cases where this leads to bugs, for example when using PropertyGrid copy / paste on a GameObject- or Component reference editor, where a detached clone will be pasted, rather than the proper reference.
- Since the `Clipboard` [requires objects to be serializable and does serialize](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.dataobject?view=netframework-4.7.1), simply removing the wrapper may not an option.
- It may be an option to instead put an object ID into the clipboard and keep an in-application lookup to resolve the ID back to the object. This could be done in a type-agnostic way, simply mapping from a randomized `long`, or counting `int` ID to an `object`.

### ToDo

- Double-check usage of both the GameObjectRefs method and similar methods for ComponentRefs and others "Refs" to make sure the API is used as a means of storing references only, rather than cloning.
- Investigate an internal ID lookup as an alternative implementation for reference storage in the clipboard. This should remain an implementation detail of the reference Set / Get extension methods for DataObject that is not visible from the outside.
- Implement if viable, find alternatives otherwise.
[...]

PR #654, opened 2018-05-11
#653 Serializable References

deanljohnson:
### Summary

Created a SerializableReferenceWrapper class to store references to objects through serialization. All that gets serialized is a long value which is used to lookup the object in a Dictionary<long, object>. Addresses #653.

### Details

All existing uses of the SetWrappedData DataObject extension method now use this wrapper except for the VectorPropertyEditor and SetIColorData. Both will end up using SerializableWrapper and a copy will be created but this maintains the current functionality as those are both value types.

### DIFF
--- a/Source/Editor/DualityEditor/Utility/SerializableReferenceWrapper.cs
+++ b/Source/Editor/DualityEditor/Utility/SerializableReferenceWrapper.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Duality.Editor.Utility
+{
+	[Serializable]
+	public class SerializableReferenceWrapper : SerializableWrapper
+	{
+		private static long NextID = 0;
+		private static readonly Dictionary<long, object> ReferenceMap 
+			= new Dictionary<long, object>();
+
+		private long ID = -1;
+
+		public sealed override object Data
+		{
+			get
+			{
+				return this.ID < 0 
+					? null 
+					: ReferenceMap[this.ID];
+			}
+			set
+			{
+				if (this.ID < 0)
+					this.ID = NextID++;
+
+				ReferenceMap[this.ID] = value;
+				base.Data = this.ID;
+			}
+		}
+
+		public SerializableReferenceWrapper(object data)
+		{
+			this.Data = data;
+		}
+	}
+}
--- a/Source/Editor/DualityEditor/Utility/SerializableReferenceWrapper.cs
+++ b/Source/Editor/DualityEditor/Utility/SerializableReferenceWrapper.cs
@@ -41,13 +41,13 @@ namespace Duality.Editor
 			object serializedObject = info.GetValue("data", typeof(long));
 			if (serializedObject is long)
 			{
-				this.ID = (long) serializedObject;
-				base.Data = referenceMap[this.ID];
+				long id = (long) serializedObject;
+				this.Data = referenceMap[id];
+				this.ID = id;
 			}
 			else
 			{
-				this.ID = -1;
-				base.Data = null;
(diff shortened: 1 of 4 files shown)


Source/Editor/DualityEditor/Utility/SerializableReferenceWrapper.cs
```diff
@@ -0,0 +1,61 @@
...
+			}
+		}
+
+		public override void GetObjectData(SerializationInfo info, StreamingContext context)
+		{
+			using (MemoryStream stream = new MemoryStream())
+			{
+				Serializer.WriteObject(this.ID, stream, typeof(BinarySerializer));
+				byte[] serializedData = stream.ToArray();
+				info.AddValue("data", serializedData);
```
ilexp: I think both serialization and deserialization here can be simplified. This setup was done in order to serialize an entire object graph using Duality serializers, but for the ID, it should be possible to just call `info.AddValue` and `info.GetValue` with the long directly.

Source/Editor/DualityEditor/Utility/SerializableReferenceWrapper.cs
```diff
@@ -0,0 +1,61 @@
...
+				return this.ID < 0 
+					? null 
+					: referenceMap[this.ID];
+			}
+			set
+			{
+				if (this.ID < 0)
+					this.ID = nextID++;
+
+				referenceMap[this.ID] = value;
```
ilexp: We shouldn't set an ID and store a reference item immediately when setting Data, but only when actually serializing it for the first time. That way we avoid cluttering the reference map with items for every `DataObject` out there that might store wrapped data.

deanljohnson:
Thanks for the review. Fixed.

Source/Editor/DualityEditor/Utility/SerializableReferenceWrapper.cs
```diff
@@ -0,0 +1,66 @@
...
+			{
+				return base.Data;
+			}
+			set
+			{
+				// Invalidate the ID in use. Could
+				// occur if this SerializableReferenceWrapper
+				// is being reused
+				this.ID = -1;
+				base.Data = value;
```
ilexp: 👍

Source/Editor/DualityEditor/Utility/SerializableReferenceWrapper.cs
```diff
@@ -0,0 +1,66 @@
...
+			object serializedObject = info.GetValue("data", typeof(long));
+			if (serializedObject is long)
+			{
+				this.ID = (long) serializedObject;
+				base.Data = referenceMap[this.ID];
+			}
+			else
+			{
+				this.ID = -1;
+				base.Data = null;
```
ilexp: `base.XY` on a property used anywhere else than inside the property override itself is a bit of an unusual pattern, and I've had to think about what would happen. Assuming that's not just me, maybe just make the field `protected` and use it directly.

It would only be a minor change though, can address this myself during merge.

deanljohnson:
Yeah I could see how the base usage was looking strange - I've fixed it.

ilexp:
Merged. Thanks!

ilexp:
Amended a commit post-merge after some testing, bullet-proofing copy / paste of references against cross application operations. Also changed serialization of ContentRefs back to use the non-reference version, since ContentRefs are actually made for serialization and translate from one application to the other, if they use similar internal content paths. Just FYI in case you wonder when looking at the code next time.

As a side note to myself, that old DataObject extension code really needs to be redone at some point. No actual design present, missed opportunities for cool stuff, application interop and more robustness.

Issue #655, opened 2018-05-12
DataObject Ref vs. Clone Data Formats


deanljohnson:
### Summary

While working on #652 we ran into an issue where copy/paste and usages of DataObject do not always agree in how stored objects should be treated (clones vs references). Investigate possible fixes to this problem.

### Details

Objects/Components/ContentRefs are currently stored in a DataObject for drag/drop, copy/paste, etc., operations. When placed into a DataObject, data is assigned a format. Currently, that format only describes the type of the data. This creates problems when a DataObject is used to hold both clones of data and references to data as those two are not equivalent but will have the same type string.

A possible solution to this would be to expand ExtMethodsDataObject to allows storing both a cloned format and a reference format. Calling code would then specify which one it was placing into the DataObject when calling SetWrappedData or on the helper methods such as SetGameObjectRefs. To make things easier on the receiving side of a DataObject, requesting a cloned format could automatically perform a deep clone of reference formatted data if a cloned format is not available. Creating a DataObject would now entail two method calls - one for cloned data and one for reference data - or we could provide a flag on the SetReferenceData method that automatically performs the DeepClone.

In general, the ExtMethodsDataObject code could use some cleanup. One thing we may be able to do is reduce the public set of methods presented in this code as the SetWrappedData/GetWrappedData/GetWrappedDataPresent methods are almost entirely used from other code in this file. The only exceptions are VectorPropertyEditor and ConversionData which both make use of those methods. The solution of using a separate data format for clones vs references will make those method signatures noisier and so it might be better to first remove their usage in outside code somehow. It seems like it may be nice to remove the current GetXPresent/Contains in place of TryGetX methods (like .Net has on Dictionary) instead as the GetX/Contains methods are doing full fetches of the data in the current implementation which could be costly.

@ilexp other issues with ExtMethodsDataObject that you see?

### TODO:
- Determine if we can reduce the public API of ExtMethodsDataObject
- Refactor and cleanup ExtMethodsDataObject where possible
- Implement new wrapper prefixes and/or replace current (current "SerializableWrapper:")
- Implement new clone format methods, paying careful attention to code reuse as this could grow out of control fairly quickly. Perhaps a new format flag to existing method signatures?
- Implement auto-conversion to clone formats from reference format on receiving end of DataObjects
- Test, test, test, as this code is used in quite a few places

ilexp:
Nice writeup, thanks!

[...]

PR #656, opened 2018-05-14
DataObject Extension Improvements

deanljohnson:
### Summary

Added support for Reference and Value data format to DataObject extension methods when using SerializedWrappers. In the process, did general cleanup and bugfixes of the ExtMethodsDataObject file and improved API in various areas. See #655.

### Details
- Added support for reference and value formatted data.
- Added unit tests for most of the code in ExtMethodsDataObject. ~75% coverage.
- Added TryGet methods and changed most code that used Contains/Get to use TryGet instead.
- Added support for automatically converting reference data stored in a DataObject to value data. DeepClones the reference data.
- Changed WrappedData API to take `IEnumerable<object>` rather than `object`
- Fixed bugs in conversion of strings to IColorData

There remain two issue I'd like to get checked.
1. Should BatchInfoPropertyEditor be using the GetBatchInfos method during a paste operation? It uses SetBatchInfos for copy, but uses ConvertOperation for paste. ConvertOperation eventually makes the same calls that GetBatchInfos would but has to go through the conversion call chain first. Is there some non-BatchInfo object that can be converted to BatchInfo?
2. Should the WrappedData methods take `object[]` rather than `IEnumerable<object>`? This keeps people from accidentally passing the wrong collection type into the SetWrappedData call and then asking for something else in GetWrappedData. ie. They pass `IEnumerable<Transform>` such as from a LINQ query and then ask for `Component[]`. If the API used `object[]` it would be more explicit and there be less of a chance that anyone would run into this problem, but it limits the API to only handling arrays rather than arbitrary collection types.

### DIFF
--- a/Source/Editor/DualityEditor/Controls/PropertyEditors/IColorDataPropertyEditor.cs
+++ b/Source/Editor/DualityEditor/Controls/PropertyEditors/IColorDataPropertyEditor.cs
@@ -178,9 +178,10 @@ namespace Duality.Editor.Controls.PropertyEditors
 			else if (e.KeyCode == Keys.V && e.Control)
 			{
 				DataObject data = Clipboard.GetDataObject() as DataObject;
-				if (data.ContainsIColorData())
+				IColorData[] colorData;
+				if (data.TryGetIColorData(out colorData))
 				{
-					this.value = data.GetIColorData<IColorData>().FirstOrDefault();
+					this.value = colorData.FirstOrDefault();
 					this.PerformSetValue();
 					this.PerformGetValue();
 					this.OnEditingFinished(FinishReason.LeapValue);
--- a/Source/Editor/DualityEditor/Controls/PropertyEditors/VectorPropertyEditor.cs
+++ b/Source/Editor/DualityEditor/Controls/PropertyEditors/VectorPropertyEditor.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Windows.Forms;
@@ -156,10 +155,11 @@ namespace Duality.Editor.Controls.PropertyEditors
 				{
 					DataObject data = Clipboard.GetDataObject() as DataObject;
 					bool success = false;
-					if (data.GetWrappedDataPresent(VectorDataFormat, DataObjectStorage.Value))
+					IEnumerable<object> storedVectors;
+					if (data.TryGetWrappedData(VectorDataFormat, DataObjectStorage.Value, out storedVectors))
 					{
-						object stored = data.GetWrappedData(VectorDataFormat, DataObjectStorage.Value).First();
-						this.SetValue(stored);
+						object vectorObj = storedVectors.First();
+						this.SetValue(vectorObj);
 						this.PerformGetValue();
 						this.OnEditingFinished(FinishReason.LeapValue);
 						this.SetFocusEditorIndex(-1, true);
--- a/Source/Editor/DualityEditor/Controls/PropertyEditors/VectorPropertyEditor.cs
+++ b/Source/Editor/DualityEditor/Controls/PropertyEditors/VectorPropertyEditor.cs
@@ -132,7 +132,7 @@ namespace Duality.Editor.Controls.PropertyEditors
 					string valString = this.editor.Select(ve => ve.Value).ToString(", ");
 					DataObject data = new DataObject();
 					data.SetText(valString);
-					data.SetWrappedData(new [] { this.DisplayedValue }, DataFormat.Value);
+					data.SetWrappedData(new [] { this.DisplayedValue }, this.DisplayedValue.GetType(), DataFormat.Value);
 
 					Clipboard.SetDataObject(data);
 					this.SetFocusEditorIndex(-1, true);
@@ -156,7 +156,7 @@ namespace Duality.Editor.Controls.PropertyEditors
 					bool success = false;
 					if (data.GetWrappedDataPresent(this.DisplayedValue.GetType(), DataFormat.Value))
 					{
-						object stored = data.GetWrappedData(this.DisplayedValue.GetType(), DataFormat.Value)[0];
+						object stored = data.GetWrappedData(this.DisplayedValue.GetType(), DataFormat.Value).First();
 						this.SetValue(stored);
 						this.PerformGetValue();
 						this.OnEditingFinished(FinishReason.LeapValue);
--- a/Source/Editor/DualityEditor/Controls/PropertyEditors/VectorPropertyEditor.cs
+++ b/Source/Editor/DualityEditor/Controls/PropertyEditors/VectorPropertyEditor.cs
@@ -132,7 +132,7 @@ namespace Duality.Editor.Controls.PropertyEditors
 					string valString = this.editor.Select(ve => ve.Value).ToString(", ");
 					DataObject data = new DataObject();
 					data.SetText(valString);
-					data.SetWrappedData(new [] { this.DisplayedValue }, this.DisplayedValue.GetType(), DataFormat.Value);
+					data.SetWrappedData(new [] { this.DisplayedValue }, DataFormat.Value);
 
 					Clipboard.SetDataObject(data);
 					this.SetFocusEditorIndex(-1, true);
--- a/Source/Editor/DualityEditor/Controls/PropertyEditors/VectorPropertyEditor.cs
+++ b/Source/Editor/DualityEditor/Controls/PropertyEditors/VectorPropertyEditor.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Windows.Forms;
@@ -131,7 +132,9 @@ namespace Duality.Editor.Controls.PropertyEditors
 					string valString = this.editor.Select(ve => ve.Value).ToString(", ");
 					DataObject data = new DataObject();
 					data.SetText(valString);
-					data.SetWrappedData(this.DisplayedValue, DataFormat.Value);
+
+					data.SetWrappedData(new [] { this.DisplayedValue }, this.DisplayedValue.GetType(), DataFormat.Value);
+
 					Clipboard.SetDataObject(data);
 					this.SetFocusEditorIndex(-1, true);
 					if (e.KeyCode == Keys.X)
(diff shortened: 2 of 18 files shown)


ilexp:
> Should BatchInfoPropertyEditor be using the GetBatchInfos method during a paste operation? It uses SetBatchInfos for copy, but uses ConvertOperation for paste. ConvertOperation eventually makes the same calls that GetBatchInfos would but has to go through the conversion call chain first. Is there some non-BatchInfo object that can be converted to BatchInfo?

This is intentional. The convert operation classes are a both convoluted and dynamic bunch, but the core idea is that the editor handles drag / drop, copy / paste and similar operations by defining at the drop / paste site what would be expected and letting the conversion system figure out if there is a way to get there. Essentially, a pathfinding operation is taking place, from the available sources to the target. If any of them succeeds, a chain of conversions is performed. 

[See here](https://github.com/AdamsLair/duality/tree/f924875f9ee24f9b4d1b557e893e08cf3dcec665/Source/Plugins/EditorBase/DataConverters) for an almost complete list of converters, but be aware that every editor plugin is free to add its own into the mix. Theoretically, retrieving a BatchInfo can happen even if there's only a Pixmap in the data, which generates or retrieves existing Texture first, Material second, and then proceeds to create a new BatchInfo based on that Material.

> Should the WrappedData methods take object[] rather than IEnumerable<object>? This keeps people from accidentally passing the wrong collection type into the SetWrappedData call and then asking for something else in GetWrappedData. ie. They pass IEnumerable<Transform> such as from a LINQ query and then ask for Component[]. If the API used object[] it would be more explicit and there be less of a chance that anyone would run into this problem, but it limits the API to only handling arrays rather than arbitrary collection types.

Hmm.. I've looked at the code and I think we're doing this slightly wrong. We're using the format string to indicate the exact type that we put in there, which requires the Get call to use the exact same type the Set call used. This stops making sense once we're putting class instances in there, where I can put in an `object`, but still get back a `Transform`, and vice versa.

Suggestion on how to solve this:

- Change `SetWrappedData` / `GetWrappedData` to not use object `Type` at all, but instead require a data format string parameter. Adjust the getter method to check for `IsAssignableFrom` relations, so it's able to return instances of `T` even when it contains `U` as long as they're related in the correct way.
- Provide two static readonly data format strings, one for Components, one for GameObjects, and use them internally for the GameObject / Component related methods.

That way we'd replace being stuck in a branch of the actual type hierarchy with defining two slots where conceptually different data (Components, GameObjects) could be.

deanljohnson:
> Change SetWrappedData / GetWrappedData to not use object Type at all, but instead require a data format string parameter. Adjust the getter method to check for IsAssignableFrom relations, so it's able to return instances of T even when it contains U as long as they're related in the correct way.

I think the idea of using a string parameter to indicate the format is a good approach and solves the issue the current design had. However, changing GetWrappedData to check for IsAssignableFrom relations feels wrong to me. The WrappedData methods and their associated SerializableWrapper classes are thin wrappings around the DataObject API which takes and returns instances of type System.Object. The methods could take and return IEnumerable<object> which removes all type handling from them (given they take the string format parameter). If you want to use these methods but get something different than `object`, you can use `OfType<T>` or `Cast<T>`. In order to check IsAssignableFrom relations they would need to have types to compare against which would require additional parameters and/or generic overloads.

The GameObject, Component, and similar extension methods are then convenience methods which handle the passing of the format string and can check IsAssignableFrom relations. Actually, the extension methods that could use them already have generic wrappers that handle this.

ilexp:
> I think the idea of using a string parameter to indicate the format is a good approach and solves the issue the current design had. However, changing GetWrappedData to check for IsAssignableFrom relations feels wrong to me. The WrappedData methods and their associated SerializableWrapper classes are thin wrappings around the DataObject API which takes and returns instances of type System.Object. The methods could take and return IEnumerable which removes all type handling from them (given they take the string format parameter). If you want to use these methods but get something different than object, you can use OfType<T> or Cast<T>. In order to check IsAssignableFrom relations they would need to have types to compare against which would require additional parameters and/or generic overloads.
> 
> The GameObject, Component, and similar extension methods are then convenience methods which handle the passing of the format string and can check IsAssignableFrom relations. Actually, the extension methods that could use them already have generic wrappers that handle this.

Agreed, good point. Let's do it this way.

> If we switch to passing in a string, that string parameter just ends up devolving into passing a type parameter, although we can gain more control over what it passed by defining some public constant strings that we hope are used.

Of course the format strings can still be used "wrong", but we should make sure the easy way is using them right. That's why I'd like to ensure users can simply use the GameObject- and Component specific methods, which use those strings correctly without exposing parameters to set them at all. The static readonly fields providing these predefined GameObject / Component slot format strings are more of an FYI for advanced users. 

Of course you can still use any other string, but then you're essentially defining a new format that exists outside the defaults and has a custom role that does not interact with the others. Could even be useful when done intentionally.

> The Component methods (and other generic methods) get around this by storing everything with the string "Component" and instead perform type checking within the component methods themselves where we have restricted the problem domain. The WrappedData methods are far too general to support the same functionality.

True, but I think that's fine. I don't see `SetWrappedData` as a general purpose "put in there whatever you want" method that gets used all the time and can solve every case - it's really just a helper so you don't have to write your own wrapper if the type you want to store doesn't happen to be .NET `[Serializable]`. 

In fact, I think any specific data type that is commonly used should eventually get its own set of methods to get and set them, potentially even equipped with the means to cast or interpret other data that matches the required format, such as the getter / setter methods for colors, which store and retrieve colors both as raw data and as strings in order to interact with text editors. I guess my point is that type / format specialized extension methods for DataObject have a bit of a translation role too, where a non-specialized method will be more limited by design.

deanljohnson:
Seems like we are in more agreement about everything than I thought :)

So, to be consistent we would also need the public static readonly(or const) strings defined for BatchInfo, ContentRefs, IColorData, and really anything else that has these extension methods and calls Set/GetWrappedData.

What do you think about maintaining both `Type` and `string` overloads for `Set/GetWrappedData` methods? Those public static strings could still be defined, but they would simply be `typeof(GameObject).FullName` or something similar. They `Type` overloads would call the `string` overload and pass `typeof(passedType).FullName`. The type overloads allow you to work at a slightly higher abstraction level whereas the string overloads give you more flexibility.

ilexp:
Okay, so I finally found at least a little bit of time to look into this PR, and I think we're almost ready to merge. I did find some remaining issues and started working on some of them myself, but would of course appreciate working together on the remaining ones.

### Progress

- Renamed `DataType` to `DataObjectStorage`, updated docs and variable names to match. `DataType` was a bit too generic and could easily be mistaken for something from serialization, cloning, or even reflection, so I tried to find something more specific.
- Removed the "allow conversion" parameter, as I can imagine no use case where it would be necessary to pass `false`. Reference-to-value conversion should be completely safe and transparent, and no other conversion can be done anyway. Reducing complexity here will probably be more useful in the long run.

### ToDo

- We have one behavior inconsistency left to fix: When passing an object as `Reference`, we'll always get a reference back. However, when passing an object as `Value`, we'll sometimes get back a serialized copy, but sometimes we'll still get back a reference - because `SerializableWrapper` only serializes when actually put into the clipboard. It doesn't trigger when I just store something in a `DataObject`, give it to someone else, and retrieve something back. An easy fix would be to move the [cloning line](https://github.com/deanljohnson/duality/blob/609bb32a87f7142d0740f179bec729e20fb1f55e/Source/Editor/DualityEditor/Utility/SerializableWrapper.cs#L63) right [into the constructor](https://github.com/deanljohnson/duality/blob/609bb32a87f7142d0740f179bec729e20fb1f55e/Source/Editor/DualityEditor/Utility/SerializableWrapper.cs#L31), and [after data re-assignment](https://github.com/deanljohnson/duality/blob/609bb32a87f7142d0740f179bec729e20fb1f55e/Source/Editor/DualityEditor/Utility/SerializableWrapper.cs#L25), but I'd like to hear your thoughts on this first.
- We definitely need to reduce the amount of low-usage code in this class, so it will be easier to maintain. Maybe we should remove all "data present" / "contains" and "get data" methods and completely move towards the `Set` / `TryGet` pattern.

deanljohnson:
> An easy fix would be to move the cloning line right into the constructor, and after data re-assignment, but I'd like to hear your thoughts on this first.

That's exactly how I would think to fix it.

> We should try to reduce the amount of low-usage code in this class, so it will be easier to maintain.

Most of it is used besides the generic component overloads and the multitude of TryGet overloads needed to handle optional parameters. If we want to reduce maintenance overhead we could probably remove many of the TryGet overloads by instead requiring all parameters to be passed explicitly. I'll go ahead and do this (if you see this after I do it and you disagree we can always revert. It's a pretty small amount of work).

> STA issues.

What do you mean tests deadlocking? I haven't had an issue running tests and I had to add `RequiresThread(ApartmentState.STA)` to the test class for AppVeyor. Still works in VS for me. Can you provide more information on this problem?

> Test Cases

Will certainly go through these with the next set of changes - most of these I had already identified as test cases and I have been trying these case as I go along.

> Obsolete API

Probably a good idea as changes made here are breaking changes. But it brings us back to the fact that this file does have a lot of code in it already.


I'll get to work on the points identified here.

deanljohnson:
Okay - I removed various TryGet overloads and generic overloads that were hardly used (some not at all). Cut the code size by about 150 lines. WrappedData, GameObject, Component, ContentRef, BatchInfo, and IColorData now all have only four methods each: Get, Set, Contains, and TryGet. While we could remove some of the Get/Contains stuff in favor of TryGet, I think the code is at a good point now of having plenty of features but without needless overloads. I'd vote to keep both Get/Contains pairs and TryGets.

I added the obsolete API to the bottom of the file and decorated them with the `Obsolete` attribute.

SerializableWrapper now always performs a clone of the data.

What remains to be addressed is the STA issue I need more information from you about and more thorough testing of all the test cases.

deanljohnson:
Because we changed how the string format parameter is determined, this is going to need significantly more testing now. All the conversion related code uses the WrappedData methods even if the DataObject was filled by SetGameObjects, SetComponents, or one of the other wrapper methods. The conversion code choose the string to check based on the type of the object being checked for - not by the same string that SetGameObjects and similar are using. We could change the strings the convenience methods use to match the type strings and that may alleviate the difficulty, but I still feel that testing has become a larger issue.

ilexp:
Good analysis and progress, nothing to add on any accounts. Also, good call on that conversion lookup 👍 I'll get back to you with the STA issue and feedback on the latest code asap.

ilexp:
### Progress

- Found a workaround on the STA deadlock issue by simulating Clipboard use instead of actual Clipboard use, so we can get around the STA requirement overall.
- SerializableWrapper and its by-ref version now expose a `IReadOnlyList<object>` rather than an `IEnumerable<object>` to allow for easy indexing and count access.
- Restructured the code flow in `DataObjectExtensions`, so all data retrieval is routed through the `TryGet` method while now considering it as the "main" method of retrieval, which is wrapped by other variants.

Also re-did the tests I mentioned earlier, everything still working as expected. With the STA issue solved and all the work you put in previously, I think we're finally good to go here. Merging this.

ilexp:
Merged to v3.

PR #659, opened 2018-05-17
Simplified Asset Importer

deanljohnson:
### Summary

Added a abstract base helper class that implements IAssetImporter and provides a simpler implementation process for new importers while reducing code duplication in existing importers.

### Details

While looking over the importer code, I noticed that the existing importers had a lot of code duplication in implementing the `IAssetImporter` interface. They all had the same process for checking file extensions, adding input to the environment, 'handling' an input, etc. 

What I did was extract a common base class from the Pixmap, Font, and AudioData importers. This base class `AssetImporter<T>` does all the common operations and has abstract methods that are implemented to perform the actual importing and exporting of data. As `AssetImporter<T>` is also generic, it allows for a more strongly typed implementation. 

Note: I did not touch ShaderAssetImporter as it can produce assets of two types (VertexShader and FragmentShader). Adding support for that would have made `AssetImporter<T>` more complicated. The goal in extracting this base class was to remove the boilerplate in the common case. If someone needs a more advanced importer, they can implement `IAssetImporter` themselves.

The other thing that the `AssetImporter<T>` class does is define a few priority 'guidelines' mirroring what `PreviewGenerator<T>` does in providing constants to base priorities off of. This way, users have an idea of what priorities they should assign to their own importers rather than just guessing at a number and seeing if it works the way they want.

The functionality of the existing importers has not changed.

### Comparison

For comparison purposes, here is an implementation of an extremely basic importer for a text resource (the text resource just has a single string field for the text files data). First, how it could be implemented using `IAssetImporter` (70 lines):

```csharp
public class TextResourceImporter : IAssetImporter
{
	private static readonly string[] sourceFileExts = new[] { ".txt" };
	protected string SourceFileExtPrimary
	{
		get { return sourceFileExts[0]; }
	}
	protected string[] SourceFileExts
	{
		get { return sourceFileExts; }
	}
	public string Id
	{
		get { return "TextResourceImporterBasic"; }
	}
	public string Name
	{
		get { return "Basic TextResource Importer"; }
	}
	public int Priority
	{
		get { return 10; }
	}
	public void PrepareImport(IAssetImportEnvironment env)
	{
		foreach (AssetImportInput input in env.HandleAllInput(this.AcceptsInput))
		{
			env.AddOutput<TextResource>(input.AssetName, input.Path);
		}
	}
	public void Import(IAssetImportEnvironment env)
	{
		foreach (AssetImportInput input in env.Input)
		{
			ContentRef<TextResource> targetRef = env.GetOutput<TextResource>(input.AssetName);

			if (targetRef.IsAvailable)
			{
				TextResource target = targetRef.Res;
				using (StreamReader reader = new StreamReader(FileOp.Open(input.Path, FileAccessMode.Read)))
					target.RawContent = reader.ReadToEnd();
				env.AddOutput(targetRef, input.Path);
			}
		}
	}
	public void PrepareExport(IAssetExportEnvironment env)
	{
		TextResource resource = env.Input as TextResource;
		if (resource != null)
		{
			env.AddOutputPath(resource.Name + this.SourceFileExtPrimary);
		}
	}
	public void Export(IAssetExportEnvironment env)
	{
		TextResource resource = env.Input as TextResource;
		string outPath = env.AddOutputPath(resource.Name + resource.RawContent);
		File.WriteAllText(outPath, resource.RawContent);
	}
	private bool AcceptsInput(AssetImportInput obj)
	{
		string ext = PathOp.GetExtension(obj.RelativePath);
		if (this.SourceFileExts.Any(e => 
			string.Equals(e, ext, StringComparison.InvariantCultureIgnoreCase)))
		{
			return true;
		}
		return false;
	}
}
```

And now a class with the exact same functionality, but using the new base class `AssetImporter<T>` (36 lines):

```csharp
public class TextResourceImporterAdvanced : AssetImporter<TextResource>
{
	private static readonly string[] sourceFileExts = new[] { ".txt" };
	protected override string SourceFileExtPrimary
	{
		get { return sourceFileExts[0]; }
	}
	protected override string[] SourceFileExts
	{
		get { return sourceFileExts; }
	}
	public override string Id
	{
		get { return "TextResourceImporterAdvanced"; }
	}
	public override string Name
	{
		get { return "Advanced TextResource Importer"; }
	}
	public override int Priority
	{
		get { return PrioritySpecialized; }
	}
	protected override void ImportResource(ContentRef<TextResource> resourceRef, AssetImportInput input, IAssetImportEnvironment env)
	{
		TextResource resource = resourceRef.Res;
		using (StreamReader reader = new StreamReader(FileOp.Open(input.Path, FileAccessMode.Read)))
			resource.RawContent = reader.ReadToEnd();
	}
	protected override void ExportResource(ContentRef<TextResource> resourceRef, string path, IAssetExportEnvironment env)
	{
		TextResource resource = resourceRef.Res;
		File.WriteAllText(path, resource.RawContent);
	}
}
```

The `AssetImporter<T>` version is simpler, is less coupled to the environment interfaces, and despite having 20 lines exactly matching that of the `IAssetImporter` (properties they both need), is half as long.


SirePi:
Nice job!
Why not take it a step further and remove also all the boilerplate property getters by adding a protected Constructor with all the required data? Something like
```csharp
protected AssetImporter(string id, string name, int priority, params string fileExtensions)
```

so that the derived classes could be even shorter
```csharp
public class TextResourceImporterAdvanced : AssetImporter<TextResource>
{
	public TextResourceImporterAdvanced
		:base("TextResourceImporterAdvanced", "Advanced TextResource Importer", PrioritySpecialized, ".txt", ".text", ".whatever")
	{ }
	
	protected override void ImportResource(ContentRef<TextResource> resourceRef, AssetImportInput input, IAssetImportEnvironment env)
	{
		TextResource resource = resourceRef.Res;
		using (StreamReader reader = new StreamReader(FileOp.Open(input.Path, FileAccessMode.Read)))
			resource.RawContent = reader.ReadToEnd();
	}
	protected override void ExportResource(ContentRef<TextResource> resourceRef, string path, IAssetExportEnvironment env)
	{
		TextResource resource = resourceRef.Res;
		File.WriteAllText(path, resource.RawContent);
	}
}
```
also, are both `Id` and `Name` really necessary?

deanljohnson:
> Why not take it a step further and remove also all the boilerplate property getters by adding a protected Constructor with all the required data? Something like:

Yeah I guess we could do that - surely cuts down on a lot of the boilerplate code. But I think it should be the following:

`protected AssetImporter(string id, string name, int priority, string primaryFileExt, params string[] fileExtensions)`

This way implementers have to pass at least the primary file extension and it then stands out from the other extensions.

> also, are both Id and Name really necessary?

Well that's part of the `IAssetImporter` interface. `Id` is stored in the resource within the editor to track which importer imported the resource and will be used during re-import operations. It really does have to be unique across all importers. The `Name` is used when the importer is displayed to the user such as when multiple importers can handle an input file and the user has to select between them. So, given the above, I'd say yes.

SirePi:
You are right on the constructor.. good catch! 😅 
As for the `Id`/`Name` point, I know it comes from the basic interface, was just wondering if it's really necessary, given that a `Name` should be unique anyway, if you still want to be able to differentiate them in case two Importers apply for the same file type. But this is definitely not important and doesn't require changes to the interface.

ilexp:
Great addition 👍 

I've looked into this and had nothing major to add review-wise, with a few minor additions I did myself before merge. Most of it is code style here and there, see the above commits for more details.

I did revert the constructor-based importer configuration in favor of the abstract property variant, which is more in line with similar classes in the Duality editor, and also makes sense when thinking about the results as per-Type statics. Didn't make it to the discussion before this was already changed - sorry about that!

I'll merge and release this asap. Thanks! 🙂

PR #661, opened 2018-05-20
Fix audio source releasing when stream is empty

LukasPirkl:
In case that audio stream contains no data (even first call to `IAudioStreamProvider.ReadStream(...)` returns false) the `NativeAudioSource` is stuck in `streamWorkerQueue`. Because of this, the source is never freed up and also the `ReadStream` method of `IAudioStreamProvider` is constantly called.

As a fix, I am setting strStopReq to EndOfStream because this is what actually happens. Then I also need to modify condition which is stopping the streaming. I tested this on sample project and also with my custom stream provider. It should handle correctly streams of any lengths (empty, shorter than initial buffers, shorter than one buffer, infinitely long).

### DIFF
--- a/Source/Platform/DefaultOpenTK/Backend/Audio/NativeAudioSource.cs
+++ b/Source/Platform/DefaultOpenTK/Backend/Audio/NativeAudioSource.cs
@@ -260,11 +260,13 @@ namespace Duality.Backend.DefaultOpenTK
 				if (this.handle == 0) return false;
 
 				ALSourceState stateTemp = AL.GetSourceState(this.handle);
-				if (stateTemp == ALSourceState.Stopped && this.strStopReq != StopRequest.None)
+				if ((stateTemp == ALSourceState.Stopped || stateTemp == ALSourceState.Initial) && this.strStopReq != StopRequest.None)
 				{
 					// Stopped due to regular EOF. If strStopReq is NOT set,
 					// the source stopped playing because it reached the end of the buffer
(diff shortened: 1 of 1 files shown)


Source/Platform/DefaultOpenTK/Backend/Audio/NativeAudioSource.cs
```diff
@@ -281,7 +283,6 @@ internal bool PerformStreaming()
 
 					// Initially play source
 					AL.SourcePlay(handle);
-					stateTemp = AL.GetSourceState(handle);
```
LukasPirkl: I noticed that this is actually not needed because stateTemp is not used after this call.

Source/Platform/DefaultOpenTK/Backend/Audio/NativeAudioSource.cs
```diff
@@ -320,6 +321,7 @@ private void PerformStreamingBegin()
 				}
 				else
 				{
+					this.strStopReq = StopRequest.EndOfStream;
 					break;
```
ilexp: I think this fix in its current form might introduce another problem: If we have a streamed, but very short audio sample, so that it fits completely into the streaming buffers at once, the initial buffer fill will `eof` and request stopping the source - so the audio will never be played, even though it would be perfectly fine to do so.

One possible solution might be to instead request stopping the stream only when triggering `eof` in the very first buffer already, so that we'd otherwise end up with a completely empty queue.

Issue #658, opened 2018-05-17
NullreferenceException in a public getter seems to throw the editor in a corrupted state


Barsonax:
### Summary

Code from pathfindax:
![image](https://user-images.githubusercontent.com/19387223/40197250-2c1e9d8e-5a13-11e8-81b3-a73c7ff7a480.png)

If a nullreference exception is thrown here the editor seems to go into a corrupted state leading to all kind of strange things like ghost components that live outside of any scene.

TODO: simplify and isolate the problem

### How to reproduce [Bugs]

- Cause nullreference exception
- TODO

ilexp:
There are no editor guards in place in order to prevent an exception from inside user code getters or setters to propagate into the editor - however, I think it should be possible to add try / catch clauses like that in spots where these kinds of errors are most likely to occur.

Is there a callstack we can work with? From where exactly is that getter called when the problem occurs?

Barsonax:
So I tried the following things:
-Make the property invisible in the UI.
-Change the property to a method.

Neither of those changes seem to solve the problem.

The code is used in a DualityApp.Mouse.ButtonDown event. Iam still in the process of isolating this completely so sorry for the current lack of info.

Barsonax:
Ah seems I have forgotten to unsubscribe the DualityApp.Mouse.ButtonDown event. This in addition to the fact that new GameObjects/Components are created when you enter/leave sandbox mode means you will be calling methods on objects that are supposed to be out of scope. So this was really a bug caused by myself.

Aside from that fact though it would be nice if the editor actually told you forgot to unsubscribe some event.

ilexp:
Let's refocus this issue towards guarding against forgotten public event subscriptions then 👍 

Can you rename / edit the issue to reflect this? Don't remove the old description and repro sample though, doesn't hurt to have them still around, maybe just "quote" them at the bottom as original text.

Barsonax:
So a solution to this would be the following structure:
`
            var list = new List<Exception>();
            for (int i = 0; i < 10; i++)
            {
                try
                {
                    SomeShutdownMethod();
                }
                catch (Exception e)
                {
                    list.Add(e);
                }
            }

            if (list.Count > 0)
            {                
                throw new AggregateException(list);
            }
`

However this has a downside since the debugger will no longer automatically break at the original exception/exceptions. This can be solved with the [DebuggerStepThrough] attribute though.

Barsonax:
[...]

PR #662, opened 2018-05-22
#658 Fix for exception during OnShutdown causing other OnShutdowns to be skipped

Barsonax:
Not yet finished and this PR is mostly for discussion of issue #658  currently.

**Possible improvements:**
- [ ] Log the individual errors in a AggregrateException as the log currently wont show the original exceptions
- [ ] Generalize and reduce the code duplication of the current implementation
- [ ] Extend this behaviour to other events


deanljohnson:
For the `ShutdownAllReversed` method - couldn't the call site just do `list.Reverse().ShutdownAll`? This is already done in a few places before `ShutdownAll` is called so there is a precedence for it. None of this code is in a hot path either so I don't think we need to worry about the bit of a performance loss.

Barsonax:
Good point I also think there might be a good possibility that this will be used for more than just shutdown (also for init, update etc?). So eventually I want to generalize the code alot more. The current implementation is more like a proof of concept to play with.

Before making the current implementation better Iam interested in what @ilexp opinion is about this topic. Maybe he had very good reasons for not doing it like this.

deanljohnson:
Well if we wanted to generalize it with the minimum amount of code duplication we could create a `TryForAll(this IEnumerable<object> objects, Action<object> action)`. It would iterate anything and perform the given actions in a try/catch and produce a AggregateException - maybe as a return value? Just an idea.

Your throw line could also check if there is only one exception in the list and throw that exception specifically if that's the case. This would be more intuitive for the user I think.

Barsonax:
> Well if we wanted to generalize it with the minimum amount of code duplication we could create a TryForAll(this IEnumerable<object> objects, Action<object> action). It would iterate anything and perform the given actions in a try/catch and produce a AggregateException - maybe as a return value? TryForAll could probably be generic as well. Just an idea.

Ye I was thinking something along this line as well:
`
[DebuggerStepThrough]
public static bool TryExecute<TObject>(this IEnumerable<TObject> objects, Action<TObject> action, out IList<Exception> exceptions)
{
	exceptions = null;
	foreach (var o in objects)
	{
		try
		{
			action.Invoke(o);
		}
		catch (Exception e)
		{
			if (exceptions == null) exceptions = new List<Exception>();
			exceptions.Add(e);
		}
	}

	return exceptions != null;
}
`

> Your throw line could also check if there is only one exception in the list and throw that exception specifically if that's the case. This would be more intuitive for the user I think.

I think that might be better handled higher up in the application by logging the individual errors inside the aggregrate exception as that wont hide any errors inside the aggregrate exception even if there are dozens (possibly with a upper limit to prevent freezing the app?).

ilexp:
> Well if we wanted to generalize it with the minimum amount of code duplication we could create a TryForAll(this IEnumerable<object> objects, Action<object> action). It would iterate anything and perform the given actions in a try/catch and produce a AggregateException - maybe as a return value? TryForAll could probably be generic as well. Just an idea.

Great idea from a design perspective, but not sure what it would do to performance, as passing-by-delegate potentially produces garbage and cannot be optimized as thoroughly as a regular function call - I'd generally try to avoid this in spots like this. If having specialized methods instead has a big impact design / maintenance wise though, we should probably consider it anyway. It's a tradeoff.

> Before making the current implementation better Iam interested in what @ilexp's opinion is about this topic. Maybe he had very good reasons for not doing it like this.

Performance considerations mostly, as a try-catch block isn't free if I recall correctly, especially when optimizations (or lack thereof) come into play. Would still be willing to consider making the change, as you're right that there's somewhat severe editor / dev bug potential when left entirely unchecked.

----

That said, this is a change that should be done based on the v3 branch, as a lot of the related code has changed, especially due to the `ICmpInitializable` split. Closing this PR, but would welcome a v3 implementation for review and gradual iteration.

Issue #244, opened 2015-11-03
Automatically expand components in Object Inspector


RockyTV:
I feel like this is a personal issue, but I'll suggest a solution in case anyone agrees with me.

I get annoyed when I load up my Duality project, and I have to expand all the components from the selected GameObject in the Object Inspector.

Perhaps there should be an option to auto expand the components, or just save their state, so if I expand one component, save, close and reopen the editor, the component is still expanded.

hsnabn:
I'm backing this. It would be a nice little feature to have.

ilexp:
Yep, sounds like a neat little improvement!

RockyTV:
I'd like to give it a go, but I can't find the code that tracks the state of the components. @ilexp mind showing the line where it defines the state?

ilexp:
You mean whether or not a Component PropertyEditor is expanded? That's actually happening in the `GroupedPropertyEditor` base class, which is part of `AdamsLair.WinForms`. Probably the easiest way to approach this would be to modify the Object Inspector to load and save its `gridExpandState` as part of its `SaveUserData` / `LoadUserData` method plus some tweaking and potentially debugging.

deanljohnson:
> Probably the easiest way to approach this would be to modify the Object Inspector to load and save its `gridExpandState` as part of its `SaveUserData` / `LoadUserData` method plus some tweaking and potentially debugging.

It's going to be a little bit more involved. Selections don't persist through a reload of the editor and so whatever SaveUserData saves will not be valid the next time the editor opens. We could delay actually setting the expand states on load until the first selection is made, but then it would have to be the same selection as before the editor closed in order for us to expand the relevant components.

As part of implementing this we could look into saving the current object selection whenever the editor closes - at which point we could then address this issue pretty simply.

Another option would be to save the expand state based on the component type somehow - so if I expand the `Transform` on one object, then the `Transform` component of any object I select will be expanded. I feel like this may a better general solution as it will in essence store the users preference for component expansion without tying it to a specific object selection and can be used throughout their use of the editor whereas this issue just handles re-opening the editor.

ilexp:
> Another option would be to save the expand state based on the component type somehow - so if I expand the Transform on one object, then the Transform component of any object I select will be expanded. I feel like this may a better general solution as it will in essence store the users preference for component expansion without tying it to a specific object selection and can be used throughout their use of the editor whereas this issue just handles re-opening the editor.

[...]

PR #663, opened 2018-05-25
#244 Expand State saved in UserData

deanljohnson:
### Summary

Addressed #244 by saving `ObjectInspectors` `ExpandState` object to UserData.

### Details

Depends on [this PR from winforms project](https://github.com/AdamsLair/winforms/pull/7).

Nothing much to say here. Along the way I also fixed a bug related to the expand state tracking. New components were not respecting the expand state:
1. Create an object (A). Add a component. 
2. Create a new object (B). Add the same component and expand it.
3. Switch back to object (A). Notice it is expanded (as expected).
4. Create a new object (C). Add the same component. Notice is is not expanded.
5. Switch to object (A). Notice the component is not expanded.

4 and 5 are where you can notice the bug. New components were always collapsed and they would then overwrite the expand state for that component.


deanljohnson:
Merged master - should be ready for review and/or merge.

PR #674, opened 2018-07-26
Coroutine implementation for v3

SirePi:
This first PR draft aims at introducing a basic-yet-complete coroutine system in Duality; following previous conversations the earlier attempts have been modified in order to follow with the new direction of self-contained scenes and object pooling whenever possible.

In addition to the available CoroutineActions, there is also a commented out one, that implements a Task as a Coroutine - not sure if it fits the scope or it's better to leave it to the user's discretion to implement something like that.


ilexp:
Joined you working on this, as it seemed more efficient than writing a convoluted review. Pretty good starting point 👍 The changes I made so far are mostly API tweaks and removals to simplify the whole coroutine deal while we figure out its first iteration.

Some things we still need to address and / or discuss, a ToDo for both of us and anyone else who might be invested in this feature:

- Support and unit-test `yield return null` - right now it looks like it would crash, but we could just interpret it as "try again next update".
- Replace `StopAction` LINQ concatenation as a complete signal with an explicit `isCompleted` bool field, plus `IsCompleted` property returning just that. Likely more efficient this way, and also more explicit / easier to understand when checking out the code.
- Disposing a coroutine should be the same as cancelling it, unless we have a specific reason not to do that.
- Wrap coroutine executions in error handling, so one coroutine crashing doesn't affect another one executing. Add an `Error` property that returns the caught exception.
- Simplify `CoroutineAction` deal, so we don't have a whole christmas tree of classes that we need to maintain and users need to be aware of. If there is any way we can simplify this to a single, sealed class that wraps a condition method / lambda without producing too much garbage, we should consider it. Need to think about this.
  - Essentially, all actions would be like `ContinuousAction`, but we'd provide some pooled, builtin ones through specialized API. 
  - For the builtin ones, we could let them implement `IEnumerator<...>` explicitly and avoid C# generating a state machine, which would allow to re-use every enumerator instance with new parameters as needed.
  - To make this work, we could allow Coroutines to be initialized with both `IEnumerator<...>` and `IEnumerable<...>` in general.
- When starting a new coroutine, should the code segment before the first yield be executed immediately, or only on the next update? Not sure on this one.
- Should we have a global `CoroutineManager` in DualityApp, in addition to the scene-local ones?

Feedback, discussion and further work appreciated! Will get back to this issue with any of these as I find time.

Source/Core/Duality/Duality.csproj
```diff
@@ -117,6 +117,9 @@
     <Compile Include="Utility\Coroutines\CoroutineAction.cs" />
     <Compile Include="Utility\Coroutines\CoroutineManager.cs" />
     <Compile Include="Utility\Log\DebugLogOutput.cs" />
+    <Compile Include="Utility\Pooling\IPoolable.cs" />
```
ilexp: `Utility\Pooling` suggests a general, framework-wide pool implementation. Let's not do this for now - additional constraints would apply and suddenly we've got X use cases all affecting how pooling needs to be designed, and the thing gets a lot more complicated. There are several pooling implementations throughout the codebase, all of which we would suddenly need to consider. 

Finding the "one size fits all" implementation would be a big scope increase here. Let's keep it all Coroutine-specific, maybe even `internal`.

ilexp: > do you mean it should also be renamed to something like CoroutinePool?

Yep, so it will be obvious to users that this is a domain-specific implementation for internal usage. We can reconsider this at some later point, but one step at a time.

SirePi:
Current status:
- [x] **Support and unit-test yield return null - right now it looks like it would crash, but we could just interpret it as "try again next update".**
- [x] **Replace StopAction LINQ concatenation as a complete signal with an explicit isCompleted bool field, plus IsCompleted property returning just that. Likely more efficient this way, and also more explicit / easier to understand when checking out the code.**
- [x] **Disposing a coroutine should be the same as cancelling it, unless we have a specific reason not to do that.**
- [ ] **Wrap coroutine executions in error handling, so one coroutine crashing doesn't affect another one executing. Add an Error property that returns the caught exception.**
Started working on this; the question is: what to do when a Coroutine crashes? Just dispose of it? or try again for another update? what if it still crashes?

- [ ] **Simplify CoroutineAction deal, so we don't have a whole christmas tree of classes that we need to maintain and users need to be aware of. If there is any way we can simplify this to a single, sealed class that wraps a condition method / lambda without producing too much garbage, we should consider it. Need to think about this.
Essentially, all actions would be like ContinuousAction, but we'd provide some pooled, builtin ones through specialized API.
For the builtin ones, we could let them implement IEnumerator<...> explicitly and avoid C# generating a state machine, which would allow to re-use every enumerator instance with new parameters as needed.
To make this work, we could allow Coroutines to be initialized with both IEnumerator<...> and IEnumerable<...> in general.
When starting a new coroutine, should the code segment before the first yield be executed immediately, or only on the next update? Not sure on this one.**
- [ ] **Should we have a global CoroutineManager in DualityApp, in addition to the scene-local ones?**
Is it useful?

ilexp:
It seems this PR was automatically closed when I merged develop-3.0 into master and deleted it afterwards. Can you re-open this PR into master?

SirePi:
Yep, since it's still a work in progress I'll get back and reopen this once the code is in a better situation.

Issue #677, opened 2018-09-03
Potential threading bug in RawList (in v3)


Barsonax:
### Summary

In v3.0 RawList uses the ReflectionHelper class for some extra optimizations. Unfortunately this ReflectionHelper class also uses a dictionary for caching internally. If RawList is used on multiple threads there is a potential for 2 or more threads to modify the dictionary at the same time

### How to reproduce

This is probably very hard to reproduce in a consistent manner. However the very fact that a dictionary is being modified from multiple threads should be enough evidence this is a potential threading bug just wating to happen. The following code will demonstrate this is happening:

```
using System;
using System.Collections.Generic;
using System.Diagnostics;
using System.Reflection;
using System.Threading;
using System.Threading.Tasks;

namespace StaticLookupTest
{
	class Program
	{
		static void Main(string[] args)
		{
			Task.Factory.StartNew(() =>
			{
				Console.WriteLine($"Value: {ReflectionHelper.StaticLookup<int>.Value}" );
			});
			Task.Factory.StartNew(() =>
			{
				Console.WriteLine($"Value: {ReflectionHelper.StaticLookup<Process>.Value}");
			});
			Task.Factory.StartNew(() =>
			{
				Console.WriteLine($"Value: {ReflectionHelper.StaticLookup<Program>.Value}");
			});
			Task.Factory.StartNew(() =>
			{
				Console.WriteLine($"Value: {ReflectionHelper.StaticLookup<float>.Value}");
			});
			Task.Factory.StartNew(() =>
			{
				Console.WriteLine($"Value: {ReflectionHelper.StaticLookup<Type>.Value}");
			});
			Task.Factory.StartNew(() =>
			{
				Console.WriteLine($"Value: {ReflectionHelper.StaticLookup<TypeInfo>.Value}");
			});
			Task.Factory.StartNew(() =>
			{
				Console.WriteLine($"Value: {ReflectionHelper.StaticLookup<decimal>.Value}");
			});
			Task.Factory.StartNew(() =>
			{
				Console.WriteLine($"Value: {ReflectionHelper.StaticLookup<bool>.Value}");
			});
			Console.ReadKey();
		}
	}

	public static class ReflectionHelper
	{
		public class StaticLookup<T>
		{			
			public static readonly int Value = Init(typeof(T).GetTypeInfo());
		}

		private static Dictionary<TypeInfo, int> _cache = new Dictionary<TypeInfo, int>();

		public static int Init(TypeInfo typeInfo)
		{
[...]

PR #679, opened 2018-09-05
Fix for threading issue in ReflectionHelper

padesso:
This is a PR addressing Issue #677.

All writes to class variables in the ReflectionHelper class have been wrapped in lock statements to compensate for potential of multiple threads concurrently modifying the collections.

### DIFF
--- a/Source/Core/Duality/Utility/ReflectionHelper.cs
+++ b/Source/Core/Duality/Utility/ReflectionHelper.cs
@@ -57,59 +57,30 @@ namespace Duality
 		/// <returns></returns>
 		public static IEnumerable<T> GetAttributesCached<T>(this MemberInfo member) where T : Attribute
 		{
-			Attribute[] result;
-			if (!customMemberAttribCache.TryGetValue(member, out result))
+			lock (customMemberAttribCache)
 			{
-				result = member.GetCustomAttributes(true).OfType<Attribute>().ToArray();
+				Attribute[] result;
+				if (!customMemberAttribCache.TryGetValue(member, out result))
+				{
+					result = member.GetCustomAttributes(true).OfType<Attribute>().ToArray();
 
-				// If it's a Type, also check implemented interfaces for (EditorHint) attributes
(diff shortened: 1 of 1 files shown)


Source/Core/Duality/Utility/ReflectionHelper.cs
```diff
@@ -121,7 +121,10 @@ private static class StaticLookup<T>
 				}
 
 				// Mind the result for later. Don't do this twice.
-				customMemberAttribCache[member] = result;
+				lock (customMemberAttribCache)
+				{
+					customMemberAttribCache[member] = result;
+				}
```
ilexp: Commenting here, but this actually counts for all similar cases in this PR as well: I believe you actually need to lock both the `TryGetValue` and the item change, since otherwise you could still change the dictionary in thread A while thread B is in the middle of `TryGetValue`.

The easiest way to go about this for now would be to just expand the `lock` block around the entire "try, else update" section. If this was highly concurrent code, it might make sense to do two separate lock statements for try and update, but I doubt it's worth the hassle in this case.

Source/Core/Duality/Utility/ReflectionHelper.cs
```diff
@@ -121,7 +121,10 @@ private static class StaticLookup<T>
 				}
 
 				// Mind the result for later. Don't do this twice.
-				customMemberAttribCache[member] = result;
+				lock (customMemberAttribCache)
```
ilexp: It's a best practice in C# to avoid locking on anything other than a `private object fooLock = new object();` created specifically for that purpose, though as long as the collections we're dealing with here are private themselves, I could live with it. Mentioning it, but I'll leave this one to you.

Source/Core/Duality/Utility/ReflectionHelper.cs
padesso: Expanded the lock to include reads from the collections, while not including any unnecessary, thread-safe code.

PR #687, opened 2018-10-03
Tilemap scroll

helle253:
Added shift+click functionality:
Allows the user to continue editing the selection rectangle in the tile palette after un-clicking. This lets the user modify the selection rectangle at will, without the need to constantly select an 'anchor point' if they want to add a row or column.

Added middle mouse scrolling:
Pressing the middle mouse button now allows the user to dynamically pan around the tile palette window.

These two additional functionalities are atomic. One cannot occur while the other is occurring and vice versa. (this also applies to just left clicking and attempting to scroll)

### DIFF
--- a/Source/Plugins/Tilemaps/Editor/Modules/SourcePaletteTilesetView.cs
+++ b/Source/Plugins/Tilemaps/Editor/Modules/SourcePaletteTilesetView.cs
@@ -16,10 +16,11 @@ namespace Duality.Editor.Plugins.Tilemaps
 		private Rectangle  selectedArea           = Rectangle.Empty;
 		private Grid<Tile> selectedTiles          = new Grid<Tile>();
 		private Point      actionBeginTilePos     = Point.Empty;
+		private bool       areTilesSelected       = false;
 		private bool       isUserSelecting        = false;
-		private bool isUserScrolling = false;
-		private int lastMouseX = -1;
-		private int lastMouseY = -1;
+		private bool       isUserScrolling        = false;
+		private int        lastMouseX             = -1;
+		private int        lastMouseY             = -1;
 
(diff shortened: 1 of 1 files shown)


helle253:
> In addition to updating the shift + click to no longer need to move, also fixed a bug that prevented the user from 'clearing' the selection rectangle via a right click.
> It should be noted that the shift + click functionality ALWAYS uses the last 'left click' anchor point, even if no rectangle is present. e.g.
> 
> 1. select a rectangle normally
> 2. right click to clear the selection rectangle
> 3. shift + click on a different spot, note that a new selection anchor point is not generated, it uses the one generated on step 1.
> 
> I don't know if this qualifies as a bug or something where either behavior would be acceptable, but I would certainly be happy to hear input.

This has been fixed. I realized shift+click prior to ANY selection rectangle used (0,0) as the anchor, which I felt wasn't appropriate.

ilexp:
Thanks for the PR! I'll leave this open for community review and will join in shortly.

helle253:
I did notice another minor bug w/ the selection rectangle in the tile palette. When resizing, if tiles within the selection are moved, the rectangle does not reset. It appears to use the first item in the next row as whatever it is pasting for the 'blank' rectangles.

This occurs in the master branch, so I don't think it's an issue introduced by me. I'll file an issue when I get a chance to record footage.

ilexp:
@mfep Thanks for reviewing! Sounds like we'll be ready for merge. 

I'll have a brief second look, hopefully soon, so we can wrap this up.

> This occurs in the master branch, so I don't think it's an issue introduced by me. I'll file an issue when I get a chance to record footage.

👍

ilexp:
Merged as per review by @mfep. Thanks everyone! 🙂

Issue #526, opened 2017-04-26
Tileset Editor draws tiles without being in focus


ChristianGreiner:
![https://gfycat.com/SpottedInfatuatedBlackandtancoonhound](https://zippy.gfycat.com/SpottedInfatuatedBlackandtancoonhound.gif)

In this gif you can see that I selected the "Auto Tile" button. After that I switched to the Tilemap Editor
and selected the selection tool. Without clicking the mouse button I navigated the mouse back to the Tileset Editor. 
When the mouse is in the view of the Tileset Editor, crazy stuff happens...

ilexp:
Fixed by @megabytefisher. Closing this.

PR #701, opened 2018-12-05
#526 Fix for Tileset Editor drawing when out of focus

megabytefisher:
This PR is a fix for issue #526. This issue is caused when the TilesetAutoTileInput instance is no longer in the editor object selection.

When a user uses auto-tile mode, the current layer's TilesetAutoTileInput instance is added to the global object selection. However, if a user then selects something else (such as a tilemap drawing tool), it can no longer retrieve the instance. This means that subsequent UI painting will draw the tile but not apply any darkening/highlighting because TilesetView_PaintTiles first checks if that instance is null.

This commit modifies the auto-tile mode so that it stores a local instance of the input object. It will still add it to the global selection when the layer is selected, but it no longer relies on value itself.

Please let me know if there are any issues with these changes.

### DIFF
--- a/Source/Plugins/Tilemaps/Editor/TilesetEditorModes/AutoTileTilesetEditorMode.cs
+++ b/Source/Plugins/Tilemaps/Editor/TilesetEditorModes/AutoTileTilesetEditorMode.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Reflection;
 using System.Linq;
@@ -198,6 +198,7 @@ namespace Duality.Editor.Plugins.Tilemaps.TilesetEditorModes
 		}
 		protected override void OnTilesetSelectionChanged(TilesetSelectionChangedEventArgs args)
 		{
+			this.currentAutoTile = null;
 			this.UpdateTreeModel();
 		}
 		protected override void OnTilesetModified(ObjectPropertyChangedEventArgs args)
@@ -230,6 +231,7 @@ namespace Duality.Editor.Plugins.Tilemaps.TilesetEditorModes
 			}
 			else
 			{
+				this.currentAutoTile = null;
 				DualityEditorApp.Deselect(this, obj => obj is TilesetAutoTileInput);
 			}
 
--- a/Source/Plugins/Tilemaps/Editor/TilesetEditorModes/AutoTileTilesetEditorMode.cs
+++ b/Source/Plugins/Tilemaps/Editor/TilesetEditorModes/AutoTileTilesetEditorMode.cs
@@ -59,12 +59,13 @@ namespace Duality.Editor.Plugins.Tilemaps.TilesetEditorModes
 		}
 
 
-		private	TreeModel        treeModel      = new TreeModel();
-		private TileConnection   hoveredArea    = TileConnection.None;
-		private bool             isUserDrawing  = false;
-		private bool             isBaseTileDraw = false;
-		private bool             isExternalDraw = false;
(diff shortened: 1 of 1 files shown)


Source/Plugins/Tilemaps/Editor/TilesetEditorModes/AutoTileTilesetEditorMode.cs
```diff
@@ -59,12 +59,13 @@ public void Invalidate()
...
-		private bool             isBaseTileDraw = false;
-		private bool             isExternalDraw = false;
-		private AutoTileDrawMode userDrawMode   = AutoTileDrawMode.Add;
+		private TilesetAutoTileInput currentAutoTile = null;
+		private	TreeModel            treeModel       = new TreeModel();
+		private TileConnection       hoveredArea     = TileConnection.None;
+		private bool                 isUserDrawing   = false;
+		private bool                 isBaseTileDraw  = false;
+		private bool                 isExternalDraw  = false;
+		private AutoTileDrawMode     userDrawMode    = AutoTileDrawMode.Add;
```
ilexp: 👍 for applying the correct tabs / spaces rules and updating alignment.

Source/Plugins/Tilemaps/Editor/TilesetEditorModes/AutoTileTilesetEditorMode.cs
```diff
@@ -227,9 +224,14 @@ protected override void OnLayerSelectionChanged(LayerSelectionChangedEventArgs a
...
 			// Update global editor selection, so an Object Inspector can pick up the AutoTile for editing
 			if (selectedNode != null)
+			{
+				this.currentAutoTile = selectedNode.AutoTileInput;
 				DualityEditorApp.Select(this, new ObjectSelection(new object[] { selectedNode.AutoTileInput }));
+			}
 			else
+			{
 				DualityEditorApp.Deselect(this, obj => obj is TilesetAutoTileInput);
+			}
```
ilexp: Why is `this.currentAutoTile` not set to null when the layer is deselected explicitly? In general, how is deselection handled, for example when leaving AutoTile mode? The main concern here is to avoid internally keeping a selection active even though the referenced object has already become invalid.

Cases to check with AutoTile mode open:

- Selecting a different Tileset in the Project View.
- Deleting the selected Tileset in the Project View.
- Removing the selected AutoTile layer from the Tileset.
- Deselecting the selected AutoTile layer explicitly.
- Selecting an AutoTile layer in the Tileset, switching to a different editing mode of the Tileset editor, selecting a different Tileset in the project view, then switching back to AutoTile mode.

Source/Plugins/Tilemaps/Editor/TilesetEditorModes/AutoTileTilesetEditorMode.cs
```diff
@@ -91,10 +92,6 @@ public override LayerEditingCaps AllowLayerEditing
 		{
 			get { return LayerEditingCaps.All; }
 		}
-		protected TilesetAutoTileInput SelectedAutoTile
-		{
-			get { return DualityEditorApp.Selection.Objects.OfType<TilesetAutoTileInput>().FirstOrDefault(); }
-		}
```
ilexp: 👍 for removing now-unused internal code.

megabytefisher:
Hi, I've made an additional commit to clear the current auto-tile input object when the user deselects the current tileset or tileset layer. I have tested all the cases you provided, and this commit fixes all issues I encountered.

Please let me know if there are any other issues.

ilexp:
Approved and merged. Thanks! 🙂

Issue #513, opened 2017-03-20
Add method to import resources in the editor via a file select dialog.


hsnabn:
### Summary

Currently, the only method of importing a resource in the Duality editor is via dragging-and-dropping a file into the Project View. Consider adding an alternative method of importing, via a file selection dialog.

#### Benefits
- Allows importing of resources in scenarios where file drag-and-drop may be unavailable.
- Makes the editor more accessible to touchpad users (drag-and-drop is kind of hard to do with a touchpad).

#### Feature Access
- Could be accessed through a tool button on the Project View.
- Might also be accessible through an option in the right-click context menu in the Project View.

#### Implementation Ideas
- The file select dialog would presumably open to the project's root folder at first launch.
- In subsequent launches, would it perhaps be possible for it to open at the folder that was last open in the dialog?

ilexp:
Implemented by @megabytefisher in PR #702. Closing this.

PR #702, opened 2018-12-07
Added ability to import via dialog in Project View

megabytefisher:
### Summary

This PR adds the ability to import files via the system file selection dialog via the Project View. It adds a context menu item and a toolbar button that will trigger the dialog. The selected files are imported to the selected folder in the project.

This feature is intended to fulfill request #513 

### Notes

The file page_white_put.png is used as the toolbar icon. It is from the famfamfam icon set.

The file ProjectFolderView.cs used tabs for alignment already, so I used that to comply with the rest of the file. I think another commit should replace the entire file's tabs.

### Screenshot

Screenshot

Let me know if there are any issues.


Source/Plugins/EditorModules/ProjectView/Modules/ProjectFolderView.cs
```diff
@@ -576,6 +588,27 @@ protected bool DisplayConfirmDeleteSelectedFiles()
 				MessageBoxIcon.Question);
 			return result == DialogResult.Yes;
 		}
+		protected void DisplayFileImportDialog()
+		{
+			OpenFileDialog dialog = new OpenFileDialog
+			{
+				Multiselect = true,
+				InitialDirectory = this.importSourcePath ?? Path.GetFullPath(DualityApp.DataDirectory)
```
deanljohnson: Please set the title of the dialog to something appropriate - maybe "Select File to Import" or something.

megabytefisher:
Alright, I've added a title. Let me know if there needs to be any other changes.

ilexp:
Merged and pushed to `release`. Should be out there in about 20 minutes.

PR #706, opened 2018-12-16
Coroutines implementation

SirePi:
This is a reworked implementation of the first, much more complex implementation (see PR #674) and takes into account all the comments to the previous version.
It's been made as a separate PR (again) to reduce the number of commits affecting the changes, and avoid silly old me merging the whole release branch..


Barsonax:
I would change the commit comment to something that explain more what the commit adds.

SirePi:
Closed for inactivity

Issue #697, opened 2018-10-23
Dedicated Rect Atlas Class


deanljohnson:
# Summary & Analysis
Duality currently uses a `List<Rect>` for image atlases to store the position of individual sprites within a Pixmap. This approach is lacking many features such as sprite pivots and sprite string tags that would be helpful for developers and necessary to implement higher level animation functionality. Consider creating an `RectAtlas` class

This issue was first discussed in #650. Below is some relevant information from that discussion:

> So there are two spots where atlases are currently used, one being `Pixmap` (pixel coordinates) and the other being `Texture` (texture / UV coordinates). Aside from differing coordinate spaces they behave exactly the same and should provide the same featureset. Right now, they're just a `List<Rect>`, but with new features to be added, it might make sense to create an actual `ImageAtlas` class that is used by both `Pixmap` and `Texture`, and exposed using their `Atlas` property.
> 
> Since atlas lookups happen a lot on a per-frame basis (every text glyph, every tile, every sprite), performance is a priority - probably not a problem, just something to keep in mind. For example, fast-path atlas indexing using an int for some internal array is a feature we might want to keep regardless of other lookup methods.
> 
> On the top of my head, I'd probably go with the following featureset for a potential `ImageAtlas`:
> 
>     * **Fast-path atlas rect retrieval via direct indexing.** Probably still via `int` that can go directly to some internal array.
> 
>     * **The possibility to tag individual rects, or groups of rects, with a string.** Could be done manually in the slicer, in the object inspector, or automatically during import when importing sprite sheets from other applications.
> 
>     * **Lookup of an index or index list via tag string.** Doesn't need to be super-fast, as it's probably more of a one-off-per-object operation.
> 
>     * **Optional definition of rect borders** for each rect side for 9-patch support, but could be used for other purposes as well.
> 
>     * **Optional definition of a rect pivot**, allowing renderers to position the sprite relative to its own position. Would grant some extra robustness for animations, but also simplify regular `SpriteRenderer` setups, so we don't need to specify the exact rect / offset for every single renderer, and adjust that everywhere when it changes at some point.
> 
> 
> ("Optional" meaning we define default values that behave as if the feature wasn't there)
> 
> With the above features, we do get some low-key animation support, but don't actually go too far into that territory. Most of the features that can be used for easing flipbook animation workflows might as well be used for different cases and are (hopefully) general purpose enough to not feel weird in a general image atlas definition.

# Proposed `RectAtlas` API
[...]

PR #708, opened 2018-12-20
Created RectAtlas class for storing Pixmap and Texture Atlas Data

deanljohnson:
### Summary

Converted `Pixmap` and `Texture` to use a new `RectAtlas` class for storing atlas information instead of `List<Rect>`. See #697 .

### Details
* Added new `RectAtlas` class
  * Implemented `IList<Rect>` so that it can be used in a similar fashion to the old implementation.
    * This should probably be removed in a future version of Duality as it doesn't add any real value, but that would be a breaking change right now.
  * Provides support for `Vector2` pivots and `string` tags per index
  * 100% unit test coverage
* Pixmap slicer updated to provide editing support for pivots and tags.
* DynamicLighting and SmoothAnimation samples custom renderers updated to use pivots defined in the `RectAtlas`.

### Kown Remaining Issues
* Samples have saved assets that are serialized with the previous `List<Rect>` implementation and therefore the Pixmaps have lost their data. This means we probably need to do two things:
  * Write in some kind of upgrade path where old serialization information is converted to the new form. Duality user code is going to have the same issues the samples do. I'm not sure how this has been handled in the past.
  * Update the assets stored with the samples to be serialized with the new data.
* In general I think this kind of change needs a pretty heavy amount of testing.

I did not implement 9-patch data storing in this PR. I think it could easily be added later on so I'd rather get this integrated/approved before we try to add more to this feature.


Source/Core/Duality/Resources/RectAtlas.cs
```diff
@@ -0,0 +1,497 @@
...
+			}
+			return false;
+		}
+
+		/// <summary>
+		/// Copies the rects in the atlas to the given array starting at the given index.
+		/// </summary>
+		/// <param name="array">The array to copy the atlas rects to</param>
+		/// <param name="arrayIndex">The index within the given array to place the atlas rects</param>
+		public void CopyTo(Rect[] array, int arrayIndex)
```
ilexp: `IList<T>` legacy: Doesn't really need to be here.

Source/Core/Duality/Resources/RectAtlas.cs
```diff
@@ -0,0 +1,497 @@
...
+			}
+		}
+
+		/// <summary>
+		/// Removes the given rect from the atlas and returns whether
+		/// or not the rect was successfully removed.
+		/// </summary>
+		/// <param name="item">The rect to remove</param>
+		/// <returns>Whether or not the given rect was removed from the atlas</returns>
+		public bool Remove(Rect item)
```
ilexp: `IList<T>` legacy: Removal by value comparison makes little sense. Instead, by-index and by-tag should be preferred.

Source/Core/Duality/Resources/RectAtlas.cs
```diff
@@ -0,0 +1,497 @@
...
+			this.RemoveAt(index);
+			return true;
+		}
+
+		/// <summary>
+		/// Returns the index of the given rect within the atlas, or -1 if the rect is not found.
+		/// </summary>
+		/// <param name="item">The rect to find the index of</param>
+		/// <returns>The index of the rect within the atlas, or -1 if the rect is not in the atlas.</returns>
+		public int IndexOf(Rect item)
```
ilexp: `IList<T>` legacy: Index search by value comparison is a niche use case that I'd argue should never happen anyway for `RectAtlas`.

Source/Core/Duality/Resources/RectAtlas.cs
```diff
@@ -0,0 +1,497 @@
...
+
+		/// <summary>
+		/// Returns all rects with the given tag or an empty enumerable if no rects
+		/// with the given tag exist. The returned rects will be in index order.
+		/// </summary>
+		/// <param name="tag">The tag of the rects to look for</param>
+		/// <returns>
+		/// The rects with the given tag or an empty enumerable if no rects with the given tag exist.
+		/// </returns>
+		public IEnumerable<Rect> GetTaggedRects(string tag)
```
ilexp: Avoid using LINQ and `IEnumerable<T>` here to avoid garbage and improve performance. Consider the following method syntax instead: `void GetTaggedItems(string tag, List<RectAtlasItem> items)`.

Source/Core/Duality/Resources/RectAtlas.cs
```diff
@@ -0,0 +1,497 @@
...
+		/// <summary>
+		/// Returns all the indices of all rects with the given tag or an empty enumerable if no rects
+		/// with the given tag exist.
+		/// </summary>
+		/// <param name="tag">The tag of the rects to look for.</param>
+		/// <returns>
+		/// The indices of rects with the given tag or an empty enumerable
+		/// if no rects with the given tag exist.
+		/// </returns>
+		public IEnumerable<int> GetTaggedIndices(string tag)
```
ilexp: Avoid using LINQ and `IEnumerable<T>` here to avoid garbage and improve performance. Consider the following method syntax instead: `void GetTaggedIndices(string tag, List<int> indices)`

Source/Core/Duality/Resources/RectAtlas.cs
```diff
@@ -0,0 +1,497 @@
...
+		/// <summary>
+		/// Sets the tag associated with the given index.
+		/// </summary>
+		/// <param name="index">The index within the atlas to find the tag for.</param>
+		/// <param name="tag">The tag to associate with the given index.</param>
+		/// <returns>The tag of the given rect atlas index</returns>
+		/// <exception cref="ArgumentOutOfRangeException">
+		/// The given index was outside the range of the number of elements in the atlas.
+		/// </exception>
+		public void SetTag(int index, string tag)
```
ilexp: Why is settings tags and pivots separate from changing rect areas? 

Assuming we have item-based indexers instead of rect-based ones (see comments above), we could scratch all specialized API and use the indexer we have instead.

Current:
```c#
public Rect this[int index] { get; set; }

public string GetTag(int index);
public void SetTag(int index, string tag);
public void SetTags(string tag, int[] indices);

public Vector2 GetPivot(int index);
public void SetPivot(int index, Vector2 pivot);
```

Item-based:
```c#
public RectAtlasItem this[int index] { get; set; }
```

Depending on the most common use cases, we can always add convenience API later on - or leave it to the user via extension methods, since the basic functionality is there.

deanljohnson:
I agree with everything you've mentioned here and that is how I'd like to do it, but doesn't that make this a heavily breaking change since `RectAtlas` would not have an API similar to the previous `List<Rect>` implementation?

Changing from `List<Rect>` to using `IList<Rect>` is still technically a breaking change, but it is easily fixed by a breaking change by any users.

ilexp:
> I agree with everything you've mentioned here and that is how I'd like to do it, but doesn't that make this a heavily breaking change since `RectAtlas` would not have an API similar to the previous `List<Rect>` implementation?
> 
> Changing from `List<Rect>` to using `IList<Rect>` is still technically a breaking change, but it is easily fixed by a breaking change by any users.

That's a good point, but under the circumstances it doesn't matter: In a commonly used area like Pixmap / Texture atlas lookups, we should avoid breaking changes in _both_ binary and source compatibility, unless we're on a major version step. `RectAtlas` replacing the old list of rects will be a binary breaking change either way - so how much it breaks the source doesn't matter going further. 

Given that, my approach here would be to find a good design first, and then see how we can integrate it without breaking stuff.

> 2. Ensure backwards compatibility to make this a non-breaking change.

As a preview of what I think could be doable here, we could

1. Replace atlas storage in Pixmap and Texture going from `List<Rect>` to `RectAtlas`, but keep the old API as-is.
1. Add new `RectAtlas` API (keep old) to Pixmap and Texture where old API used `List<Rect>`.
1. Flag the old API as `[Obsolete]`.
1. Implement a `SerializeErrorHandler` in the compatibility plugin that is able to read the old `List<Rect>` fields in Pixmap and Texture resources and convert its data to the new `RectAtlas` field.
1. In the next major version step, remove the old `[Obsolete]` API and consider renaming the new API to something more fitting or concise now that the old API is out of the way. This last item is off limits for this issue, ignore for now.

If this approach can keep full compatibility, we're completely free in how we design `RectAtlas` at the cost of having to keep a bit of obsolete API until the next major version step.

ilexp:
@deanljohnson What's the reason for closing this PR?

Issue #711, opened 2019-01-19
Upgrade Nuget.Core to 2.14


Barsonax:
### Summary

Duality is still using `Nuget.Core` 2.8.3 which does not understand the newer targetframeworks like netstandard. Consider upgrading to `Nuget.Core` 2.14 which does understand these.

In a small test during #710 I found no breaking changes and I was able to install a package.

### Analysis

- Together with #710 this should make it really easy for duality to start supporting netstandard plugins.

ilexp:
Addressed by @Barsonax. Closing this.

PR #715, opened 2019-04-03
#711 Updated nuget.core to 2.14

Barsonax:
TODO:
- [x] Update `nuget.core` to 2.14 to support netstandard (discovered in #710)
- [x] Find out why IsLatestVersion is wrongly set to true for AdamsLair.Duality.TestPlugin 1.0.0.0 in PackageManagerTests.GetLatestDualityPackages
- [x] Come up with a fix for the above problem
- [x] Verify package installs with the official sample packages work as expected.
- [x] Verify package installs with multi-framework targets work as expected (for example Singularity v0.9)
- [x] Build a binary download .zip with the new NuGet and verify both install and restore work as expected.
- [ ] After the merge, upload a new binary .zip to replace the [current one under releases](https://github.com/AdamsLair/duality/releases/tag/v3.0).

### DIFF
--- a/Source/Editor/DualityEditor/DualityEditor.csproj
+++ b/Source/Editor/DualityEditor/DualityEditor.csproj
@@ -54,9 +54,8 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="Microsoft.Build" />
-    <Reference Include="NuGet.Core, Version=2.8.50926.602, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\..\..\packages\Nuget.Core.2.8.3\lib\net40-Client\NuGet.Core.dll</HintPath>
+    <Reference Include="NuGet.Core, Version=2.14.0.832, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\..\..\packages\NuGet.Core.2.14.0\lib\net40-Client\NuGet.Core.dll</HintPath>
     </Reference>
     <Reference Include="PopupControl, Version=1.0.0.0, Culture=neutral, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
--- a/Source/Editor/DualityEditor/PackageManagement/PackageManager.cs
+++ b/Source/Editor/DualityEditor/PackageManagement/PackageManager.cs
@@ -159,7 +159,7 @@ namespace Duality.Editor.PackageManagement
 			this.manager.PackageUninstalled += this.manager_PackageUninstalled;
 			this.cache = new PackageCache(
 				this.repository,
-				repositories.OfType<LocalPackageRepository>().Any());
+				repositories.OfType<LocalPackageRepository>().Any() || repositories.OfType<LazyLocalPackageRepository>().Any());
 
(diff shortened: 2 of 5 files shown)


ilexp:
Related issue is #711, to be closed when this is merged.

ilexp:
Cleared block from issue #716.

ilexp:
What's the state of this PR, e.g. which are the ToDos that are still to be done before review? Only seeing "Test Test Test" in the main comment, not sure what exactly that means 😄

Barsonax:
There seems to be a breaking change in nuget:
image

Also VertexBatchTest.Locking fails for me locally (but not on appveyor?) not sure why.

Barsonax:
IsLatestVersion is somehow wrongly set on the 1.0.0.0 TestPlugin package:
image

Barsonax:
Modified code in PackageManager.cs a bit to catch the bug earlier in the process. Seems to have something to do with the LazyLocalPackageRepository:
image

Barsonax:
Nuget v2.8.5 is the last version that doesn't exhibit this behavior. All versions from v2.8.6 on seem to have this bug.

Barsonax:
Ah I see we have a custom piece of code that sets the is latest. This piece of code expects a LocalPackageRepository to be present. However from nuget 2.8.6 on its now wrapped in a LazyLocalPackageRepository. Fix should be easy.

Barsonax:
TODO:
- perform some real tests to verify the package manager is working as expected.

ilexp:
Great work, looking forward to the test results 👍 

> Also VertexBatchTest.Locking fails for me locally (but not on appveyor?) not sure why.

This could be related to semi-unsafe code, we should verify it's not an x86 / x64 disparity or something like that. Can you open a new issue detailing the exact failure, as well as the test init logs regarding processor architecture and process modes?

Barsonax:
Did the following tests:
- Install singularity 0.9, this is a multi framework package that targets net 4.5.2 and netstandard 2.0. It picks the net 4.5.2 version which is correct!
- Installed some other package such as tilemaps or the sample packages.

ilexp:
> @ilexp I guess you want to take a look as well before merging this?

Will do as soon as I manage.

ilexp:
## Progress

- Fixed an unrelated, but blocking bug in master branch package management.
- Debug-stepped through unit tests for packages and verified the new netstandard framework can now be parsed properly by NuGet.
- Tested package install, uninstall and restore with various official and non-official plugin and sample packages.

## ToDo

- Merge back to master, bump version numbers and trigger a new binary release.
- After the merge, upload a new binary .zip to replace the current one under releases.
- Do a final "out in the wild" test.

Looks fine to me, I'd say we're ready for merge. Will proceed when I find the next sufficiently large time frame to do so without a rush.

ilexp:
> Also VertexBatchTest.Locking fails for me locally (but not on appveyor?) not sure why.

@Barsonax Any info on this? Is this still an issue?

Barsonax:
No that test still fails. Created issue about it: https://github.com/AdamsLair/duality/issues/719

Seems to have nothing to do with this PR so lets merge this and look into that issue afterwards.

ilexp:
> Seems to have nothing to do with this PR so lets merge this and look into that issue afterwards.

Agreed. Proceeding as outlined above and as time allows 👍

ilexp:
Merged, triggered a package release, and updated the download package.

Issue #718, opened 2019-04-13
Move from the legacy nightlybuilder tool to more modern tooling


Barsonax:
### Summary

Currently duality uses the legacy nightlybuilder. Consider moving towards more modern tooling. At work iam using [NUKE build](https://nuke.build/) where you define your build in C# and can separate the build steps in targets that depend on each other. It offers full intelisense, debugging capabilities, allows you to install tools locally etc.

Example where I use it for my build:
https://github.com/Barsonax/Singularity/blob/master/_build/Build.cs

Barsonax:
@ilexp

ilexp:
Some notes on the issue:

- NightlyBuilder is a legacy tool that allowed a CI-like workflow before AppVeyor took over most of its tasks.
- By now, the only task left exclusively to NightlyBuilder is the packaging of nuspecs, and the binary download .zip for a fresh install.
- If we can automate both using regular, well-known tooling via AppVeyor, we're good to go and can basically just remove every trace of NightlyBuilder from the project.
- [The link between AppVeyor and NightlyBuilder is here](https://github.com/AdamsLair/duality/blob/7dbdba55921fcdc11593046fedf93593b08cfa55/appveyor.yml#L40), linking [this batch file](https://github.com/AdamsLair/duality/blob/a0e910f093f3f74d3b4e041665cce5ba42bd022e/Build/Scripts/Package%20Binary%20CI%20Script.bat), but only in the `release` branch, as that's the only one producing artifacts and publishing packages.

Barsonax:
The dotnet pack command can be used to generate nupkg files. However iam using the new csproj format in my project so not sure if it works with the old format which duality still uses.

ilexp:
Isn't `dotnet` part of the modern .NET Core / .NET Standard tooling? We need to check whether this produces the expected results in our case, especially since no .csproj is involved in packing here, only the .nuspecs.

Alternatively, I would assume that AppVeyor has `nuget` in their `PATH` - that could be a very straightforward replacement, if easily possible to execute in the way we need.

Barsonax:
Yes it is. It uses msbuild internally (which we could also use). Don't know how it handles the old csproj format duality uses. It works for me but iam using packagereference.

On a side note I have been using nuke to build singularity. Currently putting the more advanced stuff such as code coverage in it as well. Really liking it as its just C# and the provided libraries are nice:
[...]

PR #720, opened 2019-05-08
[WIP] Migrated projects to packagereference

Barsonax:
Migration went pretty smooth but noticed some build errors:
image

Seems it does not copy some references to the build output folder for some reason. Not sure why. I tested it with a .NET framework console app and it works correctly in that case so it might be that the portable target is causing issues here.

### DIFF
--- a/Source/Core/Duality/Duality.csproj
+++ b/Source/Core/Duality/Duality.csproj
@@ -468,12 +468,9 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
-    <Reference Include="NVorbis, Version=0.7.6.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\..\..\packages\AdamsLair.NVorbis.0.7.6\lib\portable45-net45+win8+wpa81\NVorbis.dll</HintPath>
-    </Reference>
-  </ItemGroup>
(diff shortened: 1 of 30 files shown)


ilexp:
Looked into this a bit and didn't find a way to make it work. There were two issues:

- Projects weren't able to restore packages properly, no references to the dependencies were present. This can be fixed by adding a `<NuGetTargetFramework>net45</NuGetTargetFramework>` to the project properties. After that, the packages show up directly under "References" and seem to be available in the code as well, so that's good.
- As you mentioned, referenced package binaries are not copied to the build output. I wasn't able to fix that one, and adding a `<CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>` as some internet threads suggest doesn't seem to have an effect.

Overall I'm not sure the new `PackageReference` style is fully supported for old / PCL style projects. We may need to defer this until switching to .NET Standard (see #573). As far as issue #718 goes, we might have more luck by investigating `nuget pack`-style options.

Unless you still have some options to try or want to investigate further, I'd close this PR without merge.

Barsonax:
Nope I don't have a solution for this other than switching to net standard which is another issue.

So this PR can be closed as its main purpose was to check if this was a feasible path.

PR #722, opened 2019-05-22
690 - Fixed incorrect tileset selection on resizing palette

cowmanjoe:
Very simple fix to #690. Essentially just a check whenever selected tiles are updated that they are within the bounds of the selectable tiles. If this isn't the case, the selection is deselected entirely.


ilexp:
Moved your fix to a slightly different spot to make its intentions clearer and avoid running it in situations where it may not be needed. I also removed the `areTilesSelected` field set, as I suspect that this field is actually a UI interaction temp variable judging from the surrounding code.

Looks good overall, can verify it fixes the issue 👍

Issue #698, opened 2018-10-29
Consider upgrading the C# version


Barsonax:
### Summary

Currently duality uses C# 4.0. This means we are missing the newer features that C#7.3 brings. Consider upgrading to a newer C# version.

### Analysis

- Will enable us to write cleaner more readable code. For instance C#6 has null propagation. This is a pretty long list of features so its better to actually read up on the newer C# versions than to post them here.
- We would be able to do #600 
- Will make duality incompatbile with older visual studio versions. Only applies to ppl working on the source code itself.
- Even if we don't upgrade .NETstandard will make the solution incompatible with older visual studio versions anyway.

Barsonax:
C# 8.0 among other features will bring nullable references. This is a opt in feature as the new syntax is a breaking change. It is backwards compatible once compiled though. You do need Visual studio 2019 which is currently still in preview.

Even if nullable references are not directly beneficial to duality it is to users that use duality as they can see when returns/parameters etc can be null or not. This will help users write more robust code.

It might be worth looking into once C# 8.0 is stable.

ilexp:
Alright, let's get things started. Is there a feature-by-feature list of changes to C# by language version, as well as minimum required VS toolchain and its release date? If not, can we compile one as a basis for deciding which version to upgrade to?

Barsonax:
From what I know:
VS2015 C#6
VS2017 C#7
VS2019 C#8

Since you need C#8 for nullable references I say we go for C#8. VS2019 is out already and it's stable (and ALOT faster I have to say).

Barsonax:
As a first step I say we just switch all projects to C# 8.0 and get appveyor working with it. After that we can start working on actually using the new features and changing the coding style where needed.

ilexp:
Not new info, but for the record, as I didn't see it mentioned here: We can't upgrade to C# 8.0 yet, as VS 2019 doesn't support building PCL libraries by default and the AppVeyor 2019 image doesn't have the .NET Portable Library Targeting Pack. 

So we can either do the switch to .NET Standard first (#573), or just upgrade to C# 7.3. I've created a [new discussion on the AppVeyor support](https://help.appveyor.com/discussions/problems/24595-visual-studio-2019-image-with-pcl-build-support) in the mean time.

I'll pick this up to upgrade to C# 7.3 and .NET 4.7.2.

ilexp:
## Progress

- Updated all Duality projects using the .NET Framework to 4.7.2
[...]

PR #726, opened 2019-06-17
Feature/#698 csharp80 upgrade

Barsonax:
- Enabled C# 8.0 on all projects
- Changed appveyor image to use VS2019

Note that this does not yet enable nullable references. For that the following line needs to be added to the csproj's:
```
<NullableContextOptions>enable</NullableContextOptions>
```
The above could be done in a separate PR though since its also a different issue.

EDIT:
Its now changed to:
```
<Nullable>enable</Nullable>
```

### DIFF
--- a/Samples/AudioHandling/AudioHandling.csproj
+++ b/Samples/AudioHandling/AudioHandling.csproj
@@ -34,7 +34,7 @@
     <DocumentationFile>AudioHandling.core.xml</DocumentationFile>
     <NoWarn>1591,1592,1573,1571,1570,1572</NoWarn>
     <Prefer32Bit>false</Prefer32Bit>
-    <LangVersion>4</LangVersion>
+    <LangVersion>8.0</LangVersion>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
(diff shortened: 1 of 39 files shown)


Barsonax:
Solves #698

Barsonax:
Seems we are getting a error in the appveyor build:
```
C:\projects\duality\Source\Core\Duality\Duality.csproj(479,11): error MSB4226: The imported project "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\Portable\v4.5\Microsoft.Portable.CSharp.targets" was not found
```
I believe this is because the VS2019 does not have the portable targeting pack installed.

Barsonax:
Jup portable is not even included anymore by default:
https://stackoverflow.com/questions/54861759/build-of-pcl-lib-fails-with-new-vs2019-agent

So we have 3 options:
- Modify the visual studio installation with powershell
- Don't use C# 8.0 but use C# 7.3 (no nullable references...)
- Switch to .net standard

ilexp:
I think we should take this opportunity to attempt option three and switch to .NET Standard first. If that turns out to be too massive, we can explore the other options. Let's keep this PR around for later, but continue discussion on #573.

ilexp:
Needs conflict resolve after recent `master` upgrade to .NET 4.7.2 / C# 7.3 via #698

ilexp:
Found out that we can't use C# 8.0 until we've also upgrade to .NET Core 3.0. Moved all related issues to a [separate milestone](https://github.com/AdamsLair/duality/milestone/9) so we can focus on [.NET Standard 2.0 and C# 7.3](https://github.com/AdamsLair/duality/milestone/8) right now. Closing this.

PR #741, opened 2019-08-05
Change plugin projects to netstandard and sdk style csproj's

Barsonax:
Changes:
- All pcl projects have been converted to netstandard
- Updated net framework nuspecs that still were using net45 to net472, I think we forgot to do this in the past.
- Moved most of the stuff that was in AssemblyInfo.cs to the csproj itself. With net sdk style projects the default is to put things like version and other metadata in the csproj.
- Updated the nuspecs and also added targetframework groups (this is the same as the default output when you pack a project).

Also solves #743


Samples/DynamicLighting/Editor/DynamicLighting.Editor.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
```
Barsonax: Not sure if this is actually needed?

Samples/FlapOrDie/FlapOrDie.csproj
```diff
@@ -1,75 +1,21 @@
...
+	<PropertyGroup>
+		<TargetFramework>netstandard2.0</TargetFramework>
+	</PropertyGroup>
+
   <PropertyGroup>
-    <DualityPackageExcludeParentElement />
-    <StartAction>Program</StartAction>
-    <StartProgram>$(SolutionDir)Build\Output\DualityEditor.exe</StartProgram>
-    <StartWorkingDirectory>..\</StartWorkingDirectory>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
```
Barsonax: I think this isn't needed as well. With .net sdk these properties are part of the csproj itself instead of a separate assembly.cs file

Test/Core/DualityTests.csproj
```diff
@@ -15,9 +13,10 @@
     <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\</SolutionDir>
-    <TargetFrameworkProfile />
     <RestorePackages>true</RestorePackages>
     <OutputPath>$(SolutionDir)Build\Output\</OutputPath>
+    <NuGetPackageImportStamp>
```
Barsonax: Doesnt seem to do anything

Samples/DynamicLighting/Editor/DynamicLighting.Editor.csproj
Barsonax: Turns it its not needed, compiles and runs just fine without

Test/Core/DualityTests.csproj
Barsonax: Removed this as well

Samples/FlapOrDie/FlapOrDie.csproj
Barsonax: Example: https://github.com/Barsonax/Singularity/blob/master/src/Singularity/Singularity.csproj#L6

Samples/DynamicLighting/Core/DebugObject.cs
```diff
@@ -1,60 +0,0 @@
...
-using Duality.Components;
-using Duality.Resources;
-using Duality.ColorFormat;
-using Duality.VertexFormat;
-
-namespace Debug
-{
-	[Serializable]
-	[RequiredComponent(typeof(Transform))]
-    public class DebugObject : Renderer
```
Barsonax: I believe this file was not used anymore but was still in source control. Since net sdk projects include all .cs files by default this revealed the leftover .cs file.

Barsonax:
Rebased on latest master, changes are down from 59 to 32 files :)

Samples/AudioHandling/AudioHandling.csproj
```diff
@@ -1,71 +1,21 @@
...
+		<TargetFramework>netstandard2.0</TargetFramework>
+	</PropertyGroup>
+
   <PropertyGroup>
-    <DualityPackageExcludeParentElement />
-    <StartAction>Program</StartAction>
-    <StartProgram>$(SolutionDir)Build\Output\DualityEditor.exe</StartProgram>
-    <StartWorkingDirectory>..\</StartWorkingDirectory>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
   </PropertyGroup>
```
cowmanjoe: Indentation seems inconsistent in this file, similarly in other csproj files

Source/Editor/DualityEditor/Utility/EditorHelper.cs
```diff
@@ -230,6 +231,7 @@ public static string CreateNewProject(string projName, string projFolder, Projec
 						compileElement.AddAfterSelf(newCompileElement);
 					}
 					coreProject.Save(SourceCodeProjectCorePluginFile);
+					*/
```
cowmanjoe: Why not just remove this entirely instead of commenting it out?

Issue #737, opened 2019-07-28
Consider switching to .NET SDK style projects


Barsonax:
### Summary

This is already a requirement for netstandard but we also have netframework projects that we could migrate to .NET sdk style projects. This is not strictly required but might still be preferable.

### Analysis
- Does not work for PCL's but these will be done when we migrate to netstandard
- A issue was found in #573 in the canvas unit tests when doing this. Not sure if this is the real cause though but just a heads up.
- Much cleaner csprojs, like it will fit in 1 screen with no scrolling required.
- Consistent style with the plugin projects
- dotnet tooling support! (such as easily creating nuget packages with `dotnet pack`)
- Needed for our future transition to .NET core.

Barsonax:
You can use netframework as target in a net sdk style project however stuff like WPF, winforms, web applications etc are not supported yet.

Since duality is using winforms this is a bit of a bummer. However what we can do for now is upgrade all projects that do not require winforms.

Maybe when .NET core 3.0 is out winform support in sdk style projects (for netframework) will be there too.

Barsonax:
Seems there are some more issues. Getting some failing unit tests when I changed DualityTests to be a net sdk style project. Branch can be found here:
https://github.com/Barsonax/duality/tree/feature/sdkcsprojs

We should first investigate this further before proceeding.

Barsonax:
Debugging the code both before and after this change resulted in this difference for the `RenderRedSquare` test:
Sdk style project:
![image](https://user-images.githubusercontent.com/19387223/63223960-af90e880-c1bd-11e9-9951-c2080649469c.png)

Old style project:
![image](https://user-images.githubusercontent.com/19387223/63223972-cdf6e400-c1bd-11e9-918a-f6af08c7f53d.png)

Might have forgotten something when changing the csproj I think.

Barsonax:
I think I got it figured out. Net sdk style projects output to a framework specific folder. In this case it would put the test dll in a net472 folder. When running the tests it will run in the context of this net472 folder and will only load plugins that are also in this folder. 

The tests have some implicit dependencies on certain duality plugins so they now break.

This behavior can be disabled though. However iam not sure if we want such implicit dependencies though that might be a discussion for a different issue.

Barsonax:
We could also simply add tests that test if we are running the correct backends and not some dummy graphic backend. The tests assume this but its currently not clear.

Barsonax:
[...]

PR #744, opened 2019-08-19
Changed some netframework projects to use .net sdk style csprojs

Barsonax:
Migrated our remaining netframework projects to sdk style projects.

Tests are green and the generated nuget packages seem to be working in the template. I think its also good to test a bit more with some samples.

Hope this is the last PR with thousands of changes for a while :)

Changes:
- All projects except the 2 debugging visualizer projects in the tool solution are now sdk style
- Members of some resource files are now internal

Completes #737


Barsonax:
Dud a quick test and seems that the winforms designer is now working in visual studio.

Source/Plugins/EditorBase/Properties/EditorBaseRes.Designer.cs
```diff
@@ -22,7 +22,7 @@ namespace Duality.Editor.Plugins.Base.Properties {
     [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "16.0.0.0")]
     [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
     [global::System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
-    public class EditorBaseRes {
+    internal class EditorBaseRes {
```
Barsonax, line 25: Seems the generator made this internal.

After some searching there also seems to be a public variant of the generator. However if the resources are only used internally I don't see a reason to make them public and it would actually be cleaner to make them internal.

ilexp: The reasoning here was to allow user plugins to 1:1 use the builtin resources (especially icons) when desired - not sure this was ever used in that way.

I'm fine either way here, I think both ways have their points.

ilexp:
Looks good, judging from just looking at the changes. Don't have the time for a full checkout and test right now, but maybe @SirePi can jump in, so we can ensure it's been tested on at least two machines?

Stuff to look out for (maybe verify on your machine as well @Barsonax, in case any points are missing):

- (Make sure to clean all existing output before testing)
- Does everything compile?
- Is the `Build/Output` folder populated as expected?
- When selecting a sample project as a startup project, does it run the editor as expected, with all the sample stuff working? 
  - Make sure to check one that carries both content and source code.
  - I usually go for the Tilemaps sample, followed by the DualStickSpaceShooter to tick two different complexity scenarios.
- When building nuget packages, do they end up fine, e.g. did they get the files they need?

Barsonax:
> Looks good, judging from just looking at the changes. Don't have the time for a full checkout and test right now, but maybe @SirePi can jump in, so we can ensure it's been tested on at least two machines?

@SirePi since this PR changes some csprojs visual studio can get confused. Clean everything before testing (running `git clean -fdx` after checking out and before opening visual studio should do the trick).

> Does everything compile?

 Yes

> Is the `Build/Output` folder populated as expected?

Yes, output is the same.

> When selecting a sample project as a startup project, does it run the editor as expected, with all the sample stuff working? 

No, have to check this but iam probably simply missing some entries for this in the csproj/launcherconfig.

> When building nuget packages, do they end up fine, e.g. did they get the files they need?

Since we are still using nuspecs this has not really changed alot except for some paths. I did some tests with our new template but I will check it again after fixing the previous point.

ilexp:
> So how I do it now is to copy the sample stuff over myself. This is the same as it was before right? I don't see any entries in the csproj that copies over all the content before we switched to net sdk style projects. @ilexp

Before, you could run it directly. This was achieved by (I think, it changed one or two times) setting the working directory for debug sessions to the sample folder where its `Data` and custom `Plugins` were, but launching the editor from the `Output` folder.

> This is not really a big deal in this case though as you can just set the dualityeditor project as the start project.

I'd argue that it kinda is though!

From a developer usability side, it was super easy to just run and debug any sample from the IDE, and I regularly did this to verify some core changes, using the samples as extended test cases. If it becomes an inconvenient thing to do, chances are it simply won't be done anymore at all, eliminating an in-depth test case during development. It also makes sample development more difficult if there is no longer an established way to test and develop them.

But yeah, like you said:

> Else we will run into the same problem that caused us to add a gamelauncher and gameditor project to the duality template. Maybe it will work again when we switch to .net core.

Seems we can't do anything about it though at this point - closest to getting there would be to manually change the working directory of the editor project, and explicitly selecting the editor application to run.

Maybe we can at least add a "sample launcher" project that is pre-configured to run the editor application from `Output` in the working directory of a sample? At least that way all you'd have to change is the working directory of the project, not create the entire setup. And it would document the possibility of this workflow in the repo, which might be the biggest advantage. The feature of having a different executing and working directory for the editor (or launcher) isn't otherwise apparent right now.

Barsonax:
> Maybe we can at least add a "sample launcher" project that is pre-configured to run the editor application from Output in the working directory of a sample? At least that way all you'd have to change is the working directory of the project, not create the entire setup. And it would document the possibility of this workflow in the repo, which might be the biggest advantage. The feature of having a different executing and working directory for the editor (or launcher) isn't otherwise apparent right now.

Agreed I will add something like a `SampleRunner` project

Barsonax:
Added a sample runner project that by default has its `launchSettings.json` set in such a way that it will use the DualStickSpaceShooter sample @ilexp

SirePi:
Tried it, it appears to be working properly: compiles, the runner runs.. everything looks good.
Only thing the runner's player ship rotates randomly sometimes.. but I think it's nothing to do with the PR, unless it's some indication that something else is not working as it should?

ilexp:
> EDIT: another possible option would be to make the paths that duality uses to find plugins configurable instead of hardcoded. This would allow us to keep the csproj's clean, make it possible to load multiple samples and give more freedom to endusers in how to structure their project. Not for this PR though.

It's not just about plugins, but also the `Data` folder, default application and user settings, editor data - essentially the working directory. Flexibility is always good, but not sure it would be worth the added complexity, since using a different working directory already works.

> Only thing the runner's player ship rotates randomly sometimes.. but I think it's nothing to do with the PR, unless it's some indication that something else is not working as it should?

Probably nothing to do with this PR like you said, but that still is weird. If you manage to reproduce this, maybe create a new issue to investigate. Could have something to do with input detection.

> Added a sample runner project that by default has its launchSettings.json set in such a way that it will use the DualStickSpaceShooter sample @ilexp

> Tried it, it appears to be working properly: compiles, the runner runs.. everything looks good.

Neat! Thanks 👍 Approving this PR.

Issue #749, opened 2019-09-03
Prevent null exception in "Select Type"-Form


ChristianGreiner:
### Summary

"Select Type" Form throws Null Exception when no input is given

### How to reproduce

- Open up the "Select Type" Form
- Just press Ok
![image](https://user-images.githubusercontent.com/6233308/64214908-9cb32f00-ceb2-11e9-9a83-d8e82699d4ca.png)

### Analysis

- Todo: Check if Textbox has empty value

### Attachments

```
EditorError:   ArgumentNullException: Value cannot be null.
Parameter name: key
CallStack:
   at System.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
   at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
   at Duality.GameObject.GetComponent(Type t) in C:\projects\duality\Source\Core\Duality\GameObject.cs:line 547
   at Duality.Editor.Plugins.Base.PropertyEditors.AddComponentPropertyEditor.OnAddComponentClicked() in C:\projects\duality\Source\Plugins\EditorBase\PropertyEditors\AddComponentPropertyEditor.cs:line 147
   at Duality.Editor.Plugins.Base.PropertyEditors.AddComponentPropertyEditor.OnMouseUp(MouseEventArgs e) in C:\projects\duality\Source\Plugins\EditorBase\PropertyEditors\AddComponentPropertyEditor.cs:line 110
   at AdamsLair.WinForms.PropertyEditing.GroupedPropertyEditor.OnMouseUp(MouseEventArgs e) in D:\svn\Projects\AdamsLair.WinForms\WinForms\PropertyEditing\GroupedPropertyEditor.cs:line 792
   at AdamsLair.WinForms.PropertyEditing.PropertyGrid.OnMouseUp(MouseEventArgs e) in D:\svn\Projects\AdamsLair.WinForms\WinForms\PropertyEditing\PropertyGrid.cs:line 757
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.UserControl.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

[...]

PR #751, opened 2019-09-16
Prevents null exception when selecting component type

foreverWIP:
Addresses issue #749 

Disables the OK button if no text is present in the filter. Additionally, disables the OK button if no node is currently selected from the list.

### DIFF
--- a/Source/Editor/DualityEditor/Forms/ListSelectionDialog.Designer.cs
+++ b/Source/Editor/DualityEditor/Forms/ListSelectionDialog.Designer.cs
@@ -32,160 +32,159 @@ namespace Duality.Editor.Forms
 		/// </summary>
 		private void InitializeComponent()
 		{
-            this.buttonCancel = new System.Windows.Forms.Button();
-            this.buttonOk = new System.Windows.Forms.Button();
-            this.labelInfo = new System.Windows.Forms.Label();
-            this.itemListing = new Aga.Controls.Tree.TreeViewAdv();
-            this.columnName = new Aga.Controls.Tree.TreeColumn();
-            this.columnOrigin = new Aga.Controls.Tree.TreeColumn();
-            this.nodeImage = new Aga.Controls.Tree.NodeControls.NodeStateIcon();
-            this.nodeName = new Aga.Controls.Tree.NodeControls.NodeTextBox();
-            this.nodeOrigin = new Aga.Controls.Tree.NodeControls.NodeTextBox();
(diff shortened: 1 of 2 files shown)


foreverWIP:
Taking a look at my changes, it seems the property I set up (`ShouldEnableOkButton`) doesn't just rely on text length, it also returns true when a node is selected (i.e. `itemListing.SelectedNode != null`. However, not only is the text length check unnecessary (since the filter finding a match will select that node in the list and enable the Ok button anyway) but it can also lead to a scenario where the Ok button can be enabled from a single character doesn't match the first letter of any node in the list. The current commit fixes this up.

foreverWIP:
The changes in the designer code came from me testing which events best kept the Ok button properly updated. The Designer.cs __should__ be properly rolled back now. As for the initial disabling of the Ok button, I've moved it out of the type checking so it now works on all list dialogs. I'm still uncertain about whether I could format my changes better, but functionally it's ready to go.

ilexp:
> I'm still uncertain about whether I could format my changes better, but functionally it's ready to go.

It's easier to show and tell here, so I've pushed directly to your PR branch, check out the "Style Tweaks" commit. Looked good overall, just two changes:

- Sorting of class members is generally done by member type, usually in order [static, non-static] and [field, event, property, method], and while there are exceptions we usually stick to it.
- We're careful where to place empty lines, and how many of them, since whitespace is great to provide visual cues to the reader.

Granted, the class you were working on wasn't sticking to our codebase rules 100%, so it was harder in this case to just "blend in". You can find a [full overview on the guidelines here](https://docs.duality2d.net/pages/v3/Maintainer-Guidelines.html).

-----

Another thing I changed as I tested was where the button state is updated: Since users can navigate with keyboard input only, there may not be a click event at all, but cursor key presses. Instead, I've moved the state update to the `SelectionChanged` event to cover _all_ selection changes, regardless of their source.

----

What I also notices just now is that you were adding your changes directly to `master`. It's not an issue for the Pull Request, but I strongly recommend to always do PRs with a separate `feature/xy` branch. Otherwise, rejecting or squash-merging PRs may cause problems in your own fork of the project, and it doesn't allow you to update your fork to the latest master version without also having your PR changes around. It also makes it impossible to have more than one PR at a time!

Back to topic though: Merging this 👍

ilexp:
Merged, and triggered a binary release. Your fix will be in editor version 3.1.7, should be out there in about 20 minutes. Thanks!

PR #755, opened 2019-10-01
Improved coroutines implementation

SirePi:
This is a reworked implementation of the first, much more complex implementation (see PR #674) and takes into account all the comments to the previous version.
In addition to PR #706, the capability to Reset a WaitCondition has been added, and useless backup files (unreferenced anyway) have been removed from the repository.


Source/Core/Duality/Utility/Coroutines/CoroutineManager.cs
```diff
@@ -0,0 +1,75 @@
...
+					this.lastFrameErrors.Add(c, c.LastException);
+				if (c.Status != CoroutineStatus.Running && c.Status != CoroutineStatus.Paused)
+					this.trashcan.Add(c);
+			}
+
+			foreach (Coroutine c in this.trashcan)
+			{
+				this.coroutines.Remove(c);
+				pool.ReturnOne(c);
+			}
```
deanljohnson: Couple ideas that I think would make this section more performant:

1. Currently, removing each completed coroutine from `trashcan` is O(n) to find the coroutine to delete and then O(n) to actually remove it from the list. If `trashcan` was `List<int>` storing the indexes of completed coroutines, you could iterate through `trashcan` in reverse and remove from `this.coroutines` by index (O(1) to lookup, O(n) to remove).
2. You could make the actual erasing of the coroutine from `this.coroutines` faster if it was safe to reorder coroutines by swapping completed coroutines with the last incomplete coroutine and then truncating `this.coroutines` at the end.
3. You could even go as far as to remove the use of a separate pool list and simply store the pool of coroutines in `this.coroutines` - keeping completed coroutines at the end and reusing them when necessary. Then when a coroutine completes all you would have to do is swap the completed coroutine with the last incomplete one.

2 & 3 may cause confusion if anyone using coroutines expects them to update in the same order every frame. That's a dangerous assumption for someone to make, but it could happen. It may be worth optimizing though as I remember with Unity's coroutines people were bitten by their performance quite often - we should avoid that here.

Source/Core/Duality/Utility/Coroutines/Coroutine.cs
```diff
@@ -0,0 +1,125 @@
...
+		{
+			return scene.CoroutineManager.StartNew(enumerator, name);
+		}
+
+		private IEnumerator<IWaitCondition> enumerator = null;
+
+		/// <summary>
+		/// Returns true if there are no more CoroutineActions to be performed
+		/// </summary>
+		public CoroutineStatus Status { get; private set; }
```
deanljohnson: Comment seems incorrect. I'm guessing this used to be a boolean property.

Source/Core/Duality/Utility/Coroutines/Coroutine.cs
```diff
@@ -0,0 +1,125 @@
...
+		public static Coroutine Start(Scene scene, IEnumerable<IWaitCondition> method, string name = null)
+		{
+			return Start(scene, method.GetEnumerator(), name);
+		}
+
+		/// <summary>
+		/// Starts a coroutine and registers it in the scene's CoroutineManager
+		/// </summary>
+		/// <param name="scene">The scene that will execute the coroutine</param>
+		/// <param name="method">The coroutine implementation</param>
```
deanljohnson: parameter name doesn't match constructor

Source/Core/Duality/Utility/Pooling/DerivedObjectsPool.cs
```diff
@@ -0,0 +1,52 @@
...
+			for (int i = this.poolIndex; i < this.pool.Count + this.poolIndex; i++)
+			{
+				this.searchIndex = i % this.pool.Count;
+
+				result = this.pool[this.searchIndex] as V;
+
+				if (result != null)
+				{
+					this.poolIndex = this.searchIndex;
+					this.pool.Remove(result);
```
deanljohnson: Couldn't you do `this.pool.RemoveAt(this.poolIndex)`?

Source/Core/Duality/Utility/Pooling/DerivedObjectsPool.cs
```diff
@@ -0,0 +1,52 @@
...
+
+					break;
+				}
+			}
+
+			if (result == null)
+				result = new V();
+
+			if(result is IPoolable)
+				(result as IPoolable).OnPickup();
```
deanljohnson: The compiler would hopefully optimize this, but if you do an as cast then check for null it should be faster than doing is followed by as.

Source/Core/Duality/Utility/Pooling/DerivedObjectsPool.cs
```diff
@@ -0,0 +1,52 @@
...
+
+			return result;
+		}
+
+		public void ReturnOne(T obj)
+		{
+			if (obj is IPoolable)
+				(obj as IPoolable).OnReturn();
```
deanljohnson: Recommend `as` then check for null.

Source/Core/Duality/Utility/Pooling/QueuedPool.cs
```diff
@@ -0,0 +1,37 @@
...
+		{
+			T result = null;
+
+			if (this.pool.Count > 0)
+				result = this.pool.Dequeue();
+			else
+				result = new T();
+
+			if (result is IPoolable)
+				(result as IPoolable).OnPickup();
```
deanljohnson: Recommend `as` then check for null.

Source/Core/Duality/Utility/Pooling/QueuedPool.cs
```diff
@@ -0,0 +1,37 @@
...
+
+			return result;
+		}
+
+		public void ReturnOne(T obj)
+		{
+			if (obj is IPoolable)
+				(obj as IPoolable).OnReturn();
```
deanljohnson: Recommend `as` then check for null.

SirePi:
Yeah, the code went through a bit of iterations.. might have some comments to improve 😅 
Thanks for the pointers.. regarding the is/as bits, would you go with as/if or as/elvis?

Source/Core/Duality/Utility/Coroutines/CoroutineManager.cs
SirePi: Why do you say that removing is O(n) + O(n)? the trashcan contains the same Coroutine that has to be removed from the list, no search is being done 🤔

deanljohnson:
As/if

Since you are removing by value, it’s O(n) to find the item. Then to
actually remove the item from a list, O(n) items get moved. If you wanted
to be more precise it’s really O(n/2) in both cases (amortized) but I
dropped the 1/2 term. If you remove by index, you don’t have to do the find
step as you have the index already.

Source/Core/Duality/Utility/Pooling/QueuedPool.cs
Barsonax: You can just do the is check nowadays and put the result in a variable: 'obj is IPoolable poolable'

Source/Core/Duality/Utility/Pooling/QueuedPool.cs
Barsonax: You can just do the is check nowadays and put the result in a variable: 'result is IPoolable poolable'

Source/Core/Duality/Utility/Pooling/DerivedObjectsPool.cs
Barsonax: You can just do the is check nowadays and put the result in a variable: 'result is IPoolable poolable'

Source/Core/Duality/Utility/Pooling/DerivedObjectsPool.cs
Barsonax: You can just do the is check nowadays and put the result in a variable: 'obj is IPoolable poolable'

Source/Core/Duality/Utility/Coroutines/CoroutineManager.cs
Barsonax: When you use Remove(instance) on a list it first has to search where that instance is in the list hence the O(n).

SirePi: No no, I understand where the O(n) is coming from in the Remove call, I was just wondering where the other O(n) is.. the only other loop is during the update part, but that can't be avoided.

I'll go ahead and do a removal by index for now - I'll try with a specialized pool to try and implement version 3. I think it's just safe to explicitly state that execution order of Coroutines is not guaranteed

Source/Core/Duality/Utility/Pooling/DerivedObjectsPool.cs
SirePi: Ah, my source is still on C# 4, so I couldn't use either elvis nor the inline is-variable... so is the upgrade to the newer C# features agreed on? @ilexp?

ilexp: We have [recently changed the core C# language version to C# 7.3](https://github.com/AdamsLair/duality/blob/4eb524fe5685213661069c4b1976f8781158057e/Source/Core/Duality/Duality.csproj#L30), so both are fine. I personally prefer to stick to the traditional subset of C#, but I wouldn't reject a PR for it. I might occasionally request changes in specific spots, depending on the case.

For now, please avoid elvis though - I have the suspicion that this particular syntactic sugar masks underlying issues with the code that would end up better being fixed at their root.

SirePi: All done

Source/Core/Duality/Utility/Coroutines/Coroutine.cs
SirePi: Fixed

Source/Core/Duality/Utility/Coroutines/Coroutine.cs
SirePi: Fixed

Source/Core/Duality/Utility/Coroutines/CoroutineManager.cs
SirePi: Changed to Queues to have O(1) all around the board

Barsonax:
@SirePi can you update the status on this?

SirePi:
Closing this PR. Replaced by #801

PR #756, opened 2019-10-01
Set hardware cursor capability

SirePi:
This PR allows the user to set the hardware (system) cursor used by the game window. It takes a Pixmap as base and leaves any necessary conversion to the specific graphics backend - the Point2 parameter is used to define the hotspot (active point) of the cursor.


Source/Platform/DefaultOpenTK/Backend/Graphics/NativeWindow.cs
```diff
@@ -274,5 +275,15 @@ private void OnRenderFrame(FrameEventArgs e)
 			Profile.TimeSwapBuffers.EndMeasure();
 			Profile.TimeRender.EndMeasure();
 		}
+
+		public void SetCursor(ContentRef<Pixmap> cursor, Point2 hotspot)
+		{
+			if (this.internalWindow != null && this.internalWindow.CursorVisible)
+			{
+				Pixmap pixmap = cursor.Res;
+				byte[] data = pixmap.MainLayer.Data.SelectMany(px => new byte[] { (byte)(px.B * px.A / 255), (byte)(px.G * px.A / 255), (byte)(px.R * px.A / 255), px.A }).ToArray();
```
ilexp: Can we do this without LINQ please? 😄

SirePi: Done

Source/Platform/DefaultOpenTK/Backend/Graphics/NativeWindow.cs
```diff
@@ -274,5 +275,30 @@ private void OnRenderFrame(FrameEventArgs e)
 			Profile.TimeSwapBuffers.EndMeasure();
 			Profile.TimeRender.EndMeasure();
 		}
+
+		public void SetCursor(ContentRef<Pixmap> cursor, Point2 hotspot)
+		{
+			if (!cursor.IsAvailable)
+			{
+				Logs.Core.WriteWarning("SetCursor: pixmap {0} not available", cursor.Name);
+			}
```
ilexp: I think it makes more sense here to _not_ check if `cursor` is available - hear me out:

If cursor is null, or unavailable, this is always an error, there is no way it could ever be valid input for this method. The appropriate way to deal with it would be to throw an `ArgumentNullException` in some initial validation. I'm not a fan of adding validation everywhere, _but_ when you look at the method implementation, it will actually just throw a `NullReferenceException` if `cursor` is null. Which is reasonably close. A warning on the other hand would just mask an actual error on the developer side with good intentions. To summarize:

- ❌ Warning and early-out: Unnecessarily masks a developer error.
- ✔️ ArgumentNullException: Explicit validation, ensure method is used as intended.
- ✔️ Not doing anything: Will cause an exception anyway, but without added code - personally, I'd be fine with that, though I can't argue best practices here.

-----

_Edit:_ On second thought, let's forget what I wrote above. Maybe specifying `null` should actually be valid input and mean "reset cursor to OS cursor". Can we do that?

Source/Platform/DefaultOpenTK/Backend/Graphics/NativeWindow.cs
```diff
@@ -274,5 +275,30 @@ private void OnRenderFrame(FrameEventArgs e)
...
+			else if (this.internalWindow != null && this.internalWindow.CursorVisible)
```
ilexp: What about cases where the cursor is set first, and _then_ it is made visible?

Source/Platform/DefaultOpenTK/Backend/Graphics/NativeWindow.cs
```diff
@@ -274,5 +275,30 @@ private void OnRenderFrame(FrameEventArgs e)
...
+
+				for (int i = 0, j = 0; i < pixmap.MainLayer.Data.Length; i++, j += 4)
+				{
+					ColorRgba px = pixmap.MainLayer.Data[i];
+
+					data[j] = (byte)(px.B * px.A / 255);
+					data[j + 1] = (byte)(px.G * px.A / 255);
+					data[j + 2] = (byte)(px.R * px.A / 255);
+					data[j + 3] = px.A;
+				}
```
ilexp: 👍 🙂

Source/Platform/DefaultOpenTK/Backend/Graphics/NativeWindow.cs
SirePi: Done

Source/Platform/DefaultOpenTK/Backend/Graphics/NativeWindow.cs
SirePi: Removed check, but the way the backend is currently implemented, setting SystemCursorVisible to false changes the Cursor to Empty.. not sure if it's the right approach or if it's better to work on the correct NativeWindow property...

Source/Platform/DefaultOpenTK/Backend/Graphics/NativeWindow.cs
SirePi: Yay

ilexp:
Merged. Thanks! 👍

Issue #746, opened 2019-08-26
Enable sourcelink


Barsonax:
### Summary

Sourcelink would enable consumers of duality to debug it like they had the source code. For instance here iam doing just that with [Singularity](https://github.com/Barsonax/Singularity) while its installed in a console app:
![demo](https://user-images.githubusercontent.com/19387223/63719912-4e68b500-c84e-11e9-8f71-fbcb7c31176d.gif)

https://github.com/dotnet/sourcelink

### Analysis

- No more copying over dll's and pdb's just to be able to debug duality source code from another project. Just disable just my code and you are good to go.
- Was easy to setup in a evening for Singularity.

ilexp:
Done by @mdonoughe. Closing this.

ilexp:
Unfortunately we'll have to re-open this as it turns out that SourceLink doesn't work when building packages via AppVeyor CI, rendering the recent addition inactive.

The following messages show up in the [CI build](https://ci.appveyor.com/project/AdamsLairBot/duality/builds/27890377/messages):

> Source control information is not available - the generated source link is empty.
> Unable to locate repository with working directory that contains directory 'C:\projects\duality\Source\Core\Duality'.

@mdonoughe Do you have any idea how this could be addressed?

Barsonax:
Seems to be a hardcoded path?

ilexp:
It's not a path from my machine, if that's what you're wondering - I'd guess it's just the directory where AppVeyor happens to store the project checkout.

ilexp:
Did a bit of research and I'm guessing it could be [this AppVeyor setting](https://github.com/AdamsLair/duality/blob/06b283956db6357d7c9bb01291652c1d49708795/appveyor.yml#L9):

```
# fetch repository as zip archive
shallow_clone: true
```

If there's no git, then of course SourceLink has no way of knowing where to link to. I'll try disabling this on the release branch and see if it helps.

ilexp:
That seems to have done it, the warnings are gone. I've triggered a new binary release that should now have SourceLink support, should be out there in about 30 minutes.

PR #758, opened 2019-10-02
enable sourcelink

mdonoughe:
This enables [Source Link](https://github.com/dotnet/sourcelink) information on all compatible projects in the repository. Should it only happen for projects under Source instead?

The Source Link documentation [recommends separating the PDBs into symbol packages](https://github.com/dotnet/sourcelink#alternative-pdb-distribution) instead of packaging them in the main nupkg files. That can probably be done in NightlyBuilder by passing an additional parameter to `nuget pack` and changing the PDB type if you want it.

Related to #746

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,9 @@
+<Project>
+  <!-- these settings will be applied to all projects -->
+  <PropertyGroup>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-19351-01" PrivateAssets="All"/>
+  </ItemGroup>
+</Project>


ilexp:
I haven't found the time to look into the details SourceLink myself, can you briefly elaborate on what exactly the build props file in this PR does and how it works? Especially since Duality uses hand-edited `.nuspec` files for its packages that exist outside the msbuild toolchain, I'd like to double-check with you that this works as intended.

> This enables Source Link information on all compatible projects in the repository. Should it only happen for projects under Source instead?

Doesn't hurt to have it in all projects I think.

Barsonax:
Directory.build.props will be implicitly imported I believe. Still not sure how that will work with custom nuspec files though. @mdonoughe did you test this?

mdonoughe:
When MSBuild builds a project it searches up the directory tree looking for a Directory.Build.props file and a Directory.Build.targets file. The Directory.Build.targets file is added to the beginning of the project, and the Directory.Build.targets file is added to the end.

If I delete all the .cs files in my work area and then launch the program in the debugger I am prompted about downloading the files via Source Link.

devenv_2019-10-03_19-54-05

devenv_2019-10-03_19-54-26

Source Link information is saved into the PDB so it will work as long as the user has or can get the PDBs. The handcrafted nuspec files are distributing the PDBs in the nupkg so the user should already have the PDB and it should just work. Normally now you would create separate symbol packages, but it's not required.

ilexp:
Great, thanks for the explanation and testing 👍 I'll just go ahead and merge this right away.

Issue #752, opened 2019-09-25
"Auto-Refresh of properties in Sandbox" doesn't work


ChristianGreiner:
The editor doesn't refresh properties of components, if you start the sandbox mode, even though
its enabled.
After you toggle the option, it works again.

![lul](https://giant.gfycat.com/PreciousBiodegradableHatchetfish.gif)

ilexp:
Looks like a glitch - not sure why, but potentially a small fix. Tagging this for new contributors.

ilexp:
Fixed by @SirePi. Closing this.

PR #760, opened 2019-10-05
Fix Object inspector auto refresh issue

SirePi:
This fixes issue #752. Don't know how it was managed before, but apparently `Time.MainTimer` resets every time the sandbox starts while the `lastRefreshTime` variable doesn't. This fix checks for situations in which lastRefreshTime is greater than MainTimer (possible only when MainTimer got restarted) and sets lastRefreshTime to MainTimer so that the usual logic can take place properly.


ilexp:
Ohh, that makes sense - I recently introduced that `Time` gets reset every time the sandbox is entered. Nice find. Thanks! Merging this.

Issue #731, opened 2019-07-07
Normalize Tile Palette Empty Source Tiles


ilexp:
### Summary

If there are multiple empty tiles in a Tileset, all of them can be used for painting tilemaps. While there is no visual distinction, flood-fill and runtime operations will still identify them as different tiles. This is a potential pitfall that can be removed on the editor side by normalizing tile palette selection.

### Analysis

- Similar to [how the Tile Palette normalized painted AutoTiles to the base tile](https://github.com/AdamsLair/duality/commit/039441a65f345a21c5820e9ea66e0de4db5d5519), it could normalize painted empty tiles to the default empty tile.
- Tilesets should be extended to expose a default empty tile index, which can be used for this.

ilexp:
Done by @cowmanjoe in PR #761.

PR #761, opened 2019-10-21
#731 - Use a default empty tile for all empty tiles

cowmanjoe:
Fairly simple change that adds an `int EmptyTileIndex` to the `Tileset` which will be used by the `TileMapToolSourcePalette` when making a new selection of tiles. This value is computed at `Tileset` compile time. It is set to -1 if there are no empty tiles.

### DIFF
--- a/Source/Plugins/Tilemaps/Core/Tilesets/Tileset.cs
+++ b/Source/Plugins/Tilemaps/Core/Tilesets/Tileset.cs
@@ -167,7 +167,7 @@ namespace Duality.Plugins.Tilemaps
 					this.GetCompileHashCode() != this.compileHash;
 			}
 		}
-        /// <summary>
+		/// <summary>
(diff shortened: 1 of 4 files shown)


Source/Plugins/Tilemaps/Core/Tilesets/Tileset.cs
```diff
@@ -166,7 +167,15 @@ public bool HasChangedSinceCompile
 					this.GetCompileHashCode() != this.compileHash;
 			}
 		}
-		
+        /// <summary>
```
ilexp: For some reason, this line was switched to space indentation. Please revert to tab indentation.

cowmanjoe: Oops! Fixed.

Issue #719, opened 2019-04-20
VertexBatchTest.Locking failing locally


Barsonax:
### Summary
The VertexBatchTest.Locking is failing for me locally

```
Expected: False
  But was:  True

   at Duality.Tests.Drawing.VertexBatchTest.Locking() in C:\Git\duality\Test\Core\Drawing\VertexBatchTest.cs:line 90
```

### How to reproduce
- Run the test
- It fails (well or not because it does not fail on the CI)

### Workaround
- Ignore it as it only happens locally

### Analysis
- Likely happens due to something environment specific. The question is what?

ilexp:
The specific part of the test that is failing is [this](https://github.com/AdamsLair/duality/blob/67823f4071ba2e7cca3063087bd5d8bd6c244911/Test/Core/Drawing/VertexBatchTest.cs#L84-L90):

```csharp
// Make sure that our locks released properly, i.e. allowing the array to be garbage collected
WeakReference weakRefToLockedData = new WeakReference(typedBatch.Vertices.Data);
Assert.IsTrue(weakRefToLockedData.IsAlive);
typedBatch = null;
abstractBatch = null;
GC.Collect();
Assert.IsFalse(weakRefToLockedData.IsAlive);
```

Which is dependent on garbage collection behavior, or more specifically the assumption that `GC.Collect` will free the vertex data array synchronously before returning. Could this be related? Are there ways to ensure this across runtime / GC versions?

Barsonax:
It does not do it synchronously:
https://stackoverflow.com/questions/748777/run-gc-collect-synchronously

Also I think depending on the GC like this is very error prone as it can change between runtime or config settings.

ilexp:
> Also I think depending on the GC like this is very error prone as it can change between runtime or config settings.

It is, but since this part of the test is essentially a GC test, there's not really a way around that.

> It does not do it synchronously:
> https://stackoverflow.com/questions/748777/run-gc-collect-synchronously

That's likely the issue then. Could be worth a try to use [this](https://docs.microsoft.com/en-us/dotnet/api/system.gc.collect?view=netframework-4.8#System_GC_Collect_System_Int32_System_GCCollectionMode_System_Boolean_) overload instead and also wait for pending finalizers afterwards.

ilexp:
Adjusted the test as described above. @Barsonax Can you check whether the issue is fixed on your machine?

Barsonax:
Will check when iam at my pc

[...]

PR #768, opened 2020-03-01
#719 Fixed vertex locking test failing in debug mode

Barsonax:
Moved assert code to check if the resource that a weakreference points to is alive or not after garbage collecting to a separate method. This defeats the fact that in debug mode instances might be kept alive for longer than strictly needed. Disabling inlining defeats any optimizations that might happen in release mode.

The end result is a garbage collection test that works both in debug and release.

### DIFF
--- a/Test/Core/Utility/GarbageCollectionUtils.cs
+++ b/Test/Core/Utility/GarbageCollectionUtils.cs
@@ -11,14 +11,14 @@ namespace Duality.Tests.Utility
 	public class GarbageCollectionUtils
 	{
 		/// <summary>
-		/// Asserts if the <see cref="WeakReference"/> in <see cref="CleanupTestSet"/> are collected after a garbage collection.
+		/// Asserts if the <see cref="WeakReference"/> in <see cref="GarbageCollectionTestSet"/> are collected after a garbage collection.
 		/// </summary>
 		/// <param name="func"></param>
 		// No inlining is important here to defeat any optimizations the compiler can make that could cause instances to live longer.
 		[MethodImpl(MethodImplOptions.NoInlining)]
-		public static void CheckIfCleanedUp(Func<CleanupTestSet> func)
+		public static void CheckIfCleanedUp(Func<GarbageCollectionTestSet> func)
 		{
-			CleanupTestSet cleanupTestSet = func();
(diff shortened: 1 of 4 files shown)


Test/Core/Utility/GarbageCollectionUtils.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using NUnit.Framework;
+
+namespace Duality.Tests.Utility
+{
+	public class GarbageCollectionUtils
```
ilexp, line 8: Since this is a new utility class, please add XML docs to class and at least its public members.

Test/Core/Utility/GarbageCollectionUtils.cs
```diff
@@ -0,0 +1,42 @@
...
+				GC.WaitForPendingFinalizers();
+			}
+			foreach (var item in cleanupTestSet.WeakReferences)
+			{
+				Assert.False(item.IsAlive);
+			}
+		}
+	}
+
+	public class CleanupTestSet
```
ilexp, line 29: Consider making this a private nested class, since it's an implementation detail.

Barsonax: Its used in VertexBatchTest so that won't compile

Test/Core/Utility/GarbageCollectionUtils.cs
Barsonax: added docs

Test/Core/Utility/GarbageCollectionUtils.cs
ilexp: Ah, makes sense. In that case, please move it to a separate file which is named after the class. I always struggle to find stuff and get an overview in projects where multiple types are mashed in the same file, and I doubt I'm the only one there.

I'd also suggest to (but this is on you, would approve either way) change the class name to something that suggests its specific field of use and relation to GC - maybe something like `GarbageCollectionTestSet`?

Barsonax: Yup I will

Also I agree thats a much better name :)

Issue #767, opened 2019-12-04
DrawDevice.AddVertices corrupts caller's data in picking mode


ShadowsInRain:
### Summary

When implementing custom `Duality.Components.Renderer` that caches vertex data, my objects rendered pitch-black, unless I purge cache every frame.

### How to reproduce

Cache vertex data instead of recreating it every frame.

### Workaround

Clone vertex data before sending it to `DrawDevice`. Works for me, I don't care about performance for now.

### Analysis

Documentation for `DrawDevice.AddVertices` claims that "ownership of the buffer remains at the callsite, while the IDrawDevice copies the required data into internal storage". Debugging reveals that `Color` field in my vertex data is modified after visiting this method.

It seems that in "picking mode" vertex color is being overridden in-place. I guess cloning vertex buffer before picking mode shenanigans would fix this, but not sure about performance implications.

https://github.com/AdamsLair/duality/blob/8455bd9649861e61787c24e56fe5447bafba084b/Source/Core/Duality/Drawing/DrawDevice.cs#L489

### Attachments

- [Render output example](https://i.imgur.com/YSnzi2n.png)

ilexp:
Ohh, good catch. Yep, that's a bug!

Should be a somewhat easy fix by simply applying the picking special case on the appropriate parts of the buffer _after_ copying, like you suggested. I'm adding a `Help Wanted` tag, in case someone wants to pick this up.

SirePi:
Fixed by #769

PR #769, opened 2020-03-05
Feature/issues/767

SirePi:
This patches issue #767. By just moving the array copy to before altering the vertex data and operating on the copy, you end up with no hits on performances whatsoever. 

Tested with the following custom renderer

```csharp
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading.Tasks;
using Duality;
using Duality.Components;
using Duality.Drawing;
using Duality.Resources;

namespace Output
{
	[RequiredComponent(typeof(Transform))]
	public class CustomRenderer : Component, ICmpRenderer, ICmpUpdatable
	{
		[DontSerialize] private VertexC1P3[] vertices;

		public CustomRenderer()
		{
			this.vertices = Enumerable.Range(0, 4).Select(_ => new VertexC1P3()).ToArray();

			this.vertices[0].Pos = new Vector3(Vector2.Zero * 50, 0);
			this.vertices[1].Pos = new Vector3(Vector2.UnitY * 50, 0);
			this.vertices[2].Pos = new Vector3(Vector2.One * 50, 0);
			this.vertices[3].Pos = new Vector3(Vector2.UnitX * 50, 0);

			this.vertices[0].Color = ColorRgba.White;
			this.vertices[1].Color = ColorRgba.Red;
			this.vertices[2].Color = ColorRgba.Green;
			this.vertices[3].Color = ColorRgba.Blue;
		}

		void ICmpRenderer.Draw(IDrawDevice device)
		{
			device.AddVertices<VertexC1P3>(Material.SolidWhite, VertexMode.Quads, this.vertices);
		}

		void ICmpRenderer.GetCullingInfo(out CullingInfo info)
		{
			info.Position = this.GameObj.GetComponent<Transform>().Pos;
			info.Radius = 50;
			info.Visibility = VisibilityFlag.Group0;
		}

		void ICmpUpdatable.OnUpdate()
		{
			if (DualityApp.Mouse.ButtonHit(Duality.Input.MouseButton.Left))
			{
				Camera cam = this.Scene.FindComponent<Camera>();
				cam.RenderPickingMap(DualityApp.WindowSize, DualityApp.WindowSize, false);
				ICmpRenderer renderer = cam.PickRendererAt((int)DualityApp.Mouse.Pos.X, (int)DualityApp.Mouse.Pos.Y);

				if(renderer == this)
					Logs.Game.Write("THIS!");
			}
		}
	}
}
```

(Sorry for the extra past commits..)


Source/Core/Duality/Drawing/DrawDevice.cs
```diff
@@ -519,7 +519,7 @@ public BatchInfo RentMaterial()
 			SortItem sortItem = new SortItem();
 			if (sortByDepth)
 			{
-				sortItem.SortDepth = this.CalcZSortIndex<T>(vertexBuffer, vertexCount);
+				sortItem.SortDepth = this.CalcZSortIndex<T>(slice.Data, vertexCount);
```
ilexp: Either the Z sort index needs to still be calculated based on the original vertexBuffer, or there needs to be a way to use `slice.Offset` here - otherwise the entire internal gathered vertex buffer (into which the slice refers) is used here.

Source/Core/Duality/Drawing/DrawDevice.cs
```diff
@@ -480,13 +480,17 @@ public BatchInfo RentMaterial()
 			if (vertexCount > vertexBuffer.Length) vertexCount = vertexBuffer.Length;
 			if (material == null) material = Material.Checkerboard.Res.Info;
 
+			// Move the added vertices to an internal shared buffer
+			VertexSlice<T> slice = this.drawVertices.Rent<T>(vertexCount);
+			Array.Copy(vertexBuffer, 0, slice.Data, slice.Offset, slice.Length);
+
```
ilexp: 👍  for just moving the copy a bit earlier, probably the best approach to address this.

Source/Core/Duality/Drawing/DrawDevice.cs
SirePi: Done 👍

PR #798, opened 2020-04-23
Changed and updated nvorbis to use the official version instead of ou…

Barsonax:
Fixes the warnings you get when referencing nvorbis in a netstandard project. This needs to be fixed because these warnings are transitive and also end up in endusers projects because they indirectly reference nvorbis.

So far the code compiles and unit tests are green but need to do more testing

### DIFF
--- a/Source/Core/Duality/Duality.csproj
+++ b/Source/Core/Duality/Duality.csproj
@@ -8,7 +8,7 @@
 	</PropertyGroup>
 
 	<ItemGroup>
-		<PackageReference Include="AdamsLair.NVorbis" Version="0.7.6" GeneratePathProperty="true" />
+		<PackageReference Include="NVorbis" Version="0.10.1" GeneratePathProperty="true"/>
 		<PackageReference Include="System.Threading.Tasks.Parallel" Version="4.3.0" />
 	</ItemGroup>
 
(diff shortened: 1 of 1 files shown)


Barsonax:
Audio sample seems to be working fine @ilexp maybe you have more cases we should test?

ilexp:
Check out the DualStickSpaceShooter example for an (at least somewhat) real-world audio test. Make sure to put in at least a bit of playtime, so any potential bugs get a chance to occur. Things to check:

- Play in launcher, maybe once from start to end to ensure a bit of playtime.
- Play in editor, then stop the sandbox mode while a bit of background audio is playing, then play again, then stop the sandbox mode again. Maybe repeat 2-3 times.

Something like that. If all looks well, I guess it's worth a shot. The only custom change I found glancing over the commits was [this one](https://github.com/AdamsLair/nvorbis/commit/75c2bfde3d8d2180e6a34c53e23210fad576eb22), but I'd hope that's history by now.

Barsonax:
I keep dying so didn't finish the game...

What I noticed is that the sound keeps playing when I pause the game. However this also happens on master for me. Doesn't sound right to be but maybe this was always the case already?

Other than that seems to work just fine.

ilexp:
> I keep dying so didn't finish the game...

It does sound like you still got some playtime though 😄 Should be fine then. Audio doesn't change considerably towards the end.

> What I noticed is that the sound keeps playing when I pause the game. However this also happens on master for me. Doesn't sound right to be but maybe this was always the case already?

That is expected behavior, yep.

> Other than that seems to work just fine.

No objections then - I'm fine with trying this, just be vigilant in case any audio-related stuff is reported.

Issue #787, opened 2020-04-10
Warning when referencing opentk nuget


Barsonax:
### Summary

When adding `AdamsLair.Duality.Backend.DefaultOpenTK` to your csproj you will get the following warning message:
```
MSB3246	Resolved file has a bad image, no metadata, or is otherwise inaccessible. Could not load file or assembly 'OpenALSoft64.dll' or one of its dependencies. An attempt was made to load a program with an incorrect format.	
```

This seems to be caused because this is a native dll and thus needs to be in a different folder in the nuget package
https://stackoverflow.com/questions/48122173/warning-when-using-nuget-package-that-contains-native-dll

Barsonax:
So I tried using the runtimes folder for the native openal dlls but couldn't get it to work. Documentation is quite scarce so I think Ill just let it rest for now since it only gives a warning.

Barsonax:
Since we know this does not break anything we might consider adding a ignore for this particular error

PR #812, opened 2020-04-30
Fixed opentk Resolved file has a bad image warning

Barsonax:
Moved the openalsoft libraries to the runtimes\any\native folder in the nuget package

In the template I had to add `<RuntimeIdentifier>any</RuntimeIdentifier>` to the csproj so that it knows to pick up the any folder.

This solves the warning you otherwise get because it was trying to load the native dll's as a non native dll.

Solves #787

### DIFF
--- a/Build/NuGetPackageSpecs/DefaultOpenTKBackend.Core.nuspec
+++ b/Build/NuGetPackageSpecs/DefaultOpenTKBackend.Core.nuspec
@@ -25,7 +25,7 @@
   <files>
     <file src="..\Output\Plugins\DefaultOpenTK.Core.dll" target="lib\net472" />
     <file src="..\Output\Plugins\DefaultOpenTK.Core.pdb" target="lib\net472" />
-    <file src="..\Output\OpenALSoft32.dll" target="lib\net472" />
-    <file src="..\Output\OpenALSoft64.dll" target="lib\net472" />
+    <file src="..\Output\OpenALSoft32.dll" target="runtimes\any\native" />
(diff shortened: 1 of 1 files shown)


Barsonax:
Seems that this library is windows only though? @ilexp do you have to use something different on linux?

ilexp:
OpenALSoft is a fallback for Windows systems that do not have any OpenAL drivers installed, and they remain inactive if OpenAL comes with the used audio drivers anyway. On Linux / MacOS, Duality does not bring along a fallback, but in the same way it's possible that these systems do have OpenAL drivers present in some form themselves. It's not a strict requirement for OpenAL audio support, but it increases compatibility on Windows.

Barsonax:
I was asking because I now put the dll's in the any folder. This works but if they are only intended for windows then thats not exactly correct. Would be cleaner to put them in win-x64 and win-x84 folders I think. I believe you can put in multiple `RuntimeIdentifier` but have to check how that will exactly behave.

ilexp:
I guess in that case we're in a Windows only context anyway, since package installs are only relevant to the developer environment setup, not deployed games. So the package would currently only be installed in a Windows setup, or one that doesn't care about those files. It might not be clean, but shouldn't cause issues. 

However: When the editor one day runs on non-Windows platforms, it might actually be correct to include the files there, since you could theoretically develop a game on Linux or Mac, which you plan to deploy on Windows, so the files would need to be there at least for deployment / build purposes.

That's way ahead into the future though. Point is, it might not be all that wrong, depends on how you look at it. I think either way would be fine for now.

Barsonax:
Hmm Ill keep it this way for now. Not really relevant atm.

PR #819, opened 2020-05-13
Fixed margin drawing in TilesetCompiler

Seti-0:


ilexp:
@Barsonax @SirePi Do you have time to do a quick review by testing the change locally? Maybe run the tilemaps sample and import a second non-square tilemap, verify that both work as expected. If not, I can take this one soon-ish.

When this is merged, we can do a v3.x package release with the fix by running the package update script and merging into `release`, which I'd leave to you guys.

SirePi:
Looks good 👍

SirePi:
@ilexp well, I arrived here, but now I'm stuck at the update version bit, as I can't update because I need to change the path for git in the updater, but if I do it complains that I can't update with uncommitted changes, which I don't want to commit 😂

ilexp:
@AdamsLair/duality-core Managed to fix it, here's a summary FYI:

As it turns out, I branched off `archive/v3.x` a little too early - there was another package release after that (Duality v3.2.1), so the `release` branch was more up to date than the archive branch. All was fine, but when the package update script ran, it of course used the old dependency versions from the archive branch. After merging into release, those versions mismatched with the actual versions on that branch. End result: No packages were released.

I fixed this by [bringing the v3.x branch up-to-date](https://github.com/AdamsLair/duality/commit/6ccd55cc41768211ccce48602efb064b1cf6e48b), [rerunning the version updater script](https://github.com/AdamsLair/duality/commit/04a64eed2fcd846de720917e77fa47facce04a4a) and [re-merging into release](https://github.com/AdamsLair/duality/commit/a9f7e23fa22174f6707ad4d5a450bcf167ee312e). The packages should now be on their way, and the issue should not happen again in the future.

@Seti-0 None of this had anything to do with your contribution by the way. Thanks, and congrats to your first merged PR! 🙂

Issue #817, opened 2020-05-12
Coroutines evolution


SirePi:
### Summary
This thread is to continue the discussions going on in the Coroutines' implementation PR #801, and to avoid cluttering the PR itself.

In particular, it emerged that further down the road, it could be worthy to investigate the following two aspects that are currently not managed,  related to the management aspects of the new Coroutines system.

### Saving/Loading Coroutines
As it is now, the system does not support saving and loading of Coroutines (it might be possible that during de/serialization the status is preserved "automagically", were the `coroutineManager` be serializable in the Scene, but testing is definitely required to be sure of the results of this operation.
From a quick glance at Unity's references, it appears that this topic still doesn't have an out-of-the-box solution, and relies on the user's management to determine which variables need to be saved and how to restore the Coroutine to the desired state.
https://docs.unity3d.com/Manual/Coroutines.html
https://docs.unity3d.com/ScriptReference/Coroutine.html
https://answers.unity.com/questions/1433878/how-to-store-the-state-of-ienumertor-and-resume-th.html
https://forum.unity.com/threads/how-can-i-save-and-load-a-coroutine-state.796401/

### Automatically clearing up Coroutines when the spawning `GameObject` gets removed
Currently the system doesn't track, and is not interested in, the context in which a Coroutine has been spawned; the Coroutine has the capability to affect any number of Entities in any Scene, and its execution is anyway managed so that an error would not result in an unrecoverable state of the engine.

 A possible solution, as proposed by @Barsonax, would be to assign to each `GameObject` (and/or `Component`) a list of Coroutines that would be automatically Cancelled once their respective "container" is being removed from the Scene. (Bonus: they could be paused/resumed automatically on de/activation). This would be an extra facility, in addition to the current way to start Coroutines.

Barsonax:
> A possible solution, as proposed by @Barsonax, would be to assign to each GameObject (and/or Component) a list of Coroutines that would be automatically Cancelled once their respective "container" is being removed from the Scene. (Bonus: they could be paused/resumed automatically on de/activation). This would be an extra facility, in addition to the current way to start Coroutines.

I think we should do this on component level and generalize it so it will work for more than just coroutines. In the future we might also have a job system which makes use of this and the enduser might also want to use this API.

Barsonax:
Another thing we might want to look into in the future:
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.iasyncenumerable-1?view=dotnet-plat-ext-3.1

[...]

PR #821, opened 2020-05-19
Implementation of WaitUntil.CoroutineEnds

SirePi:
This is an implementation of `WaitUntil.CoroutineEnds` as described [here](https://github.com/AdamsLair/duality/issues/817#issuecomment-630703411).

### DIFF
--- a/Source/Core/Duality/Utility/Coroutines/WaitUntil.cs
+++ b/Source/Core/Duality/Utility/Coroutines/WaitUntil.cs
@@ -13,7 +13,8 @@ namespace Duality.Utility.Coroutines
 		{
 			Frames,
 			GameTime,
-			RealTime
+			RealTime,
+			Coroutine
 		}
 
 		/// <summary>
@@ -21,8 +22,9 @@ namespace Duality.Utility.Coroutines
 		/// </summary>
 		public static readonly WaitUntil NextFrame = new WaitUntil(1, WaitType.Frames);
 
-		private readonly WaitType type;
 		private float internalValue;
+		private readonly Func<bool> subroutineMoveNext;
+		private readonly WaitType type;
 
 		public bool IsComplete
 		{
@@ -32,7 +34,15 @@ namespace Duality.Utility.Coroutines
 		private WaitUntil(float startingValue, WaitType type)
 		{
 			this.internalValue = startingValue;
+			this.subroutineMoveNext = null;
 			this.type = type;
+		}
+
+		private WaitUntil(Func<bool> subroutineMoveNext)
+		{
+			this.internalValue = float.PositiveInfinity;
+			this.subroutineMoveNext = subroutineMoveNext;
+			this.type = WaitType.Coroutine;
 		}
 
 		internal void Update()
@@ -49,6 +59,10 @@ namespace Duality.Utility.Coroutines
 
 				case WaitType.RealTime:
 					this.internalValue -= Time.UnscaledDeltaTime;
+					break;
+
+				case WaitType.Coroutine:
+					this.internalValue = this.subroutineMoveNext() ? float.PositiveInfinity : 0;
 					break;
 			}
 		}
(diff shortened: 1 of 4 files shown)


Source/Core/Duality/Utility/Coroutines/WaitUntil.cs
```diff
@@ -13,16 +13,18 @@ private enum WaitType
...
 		}
 
 		/// <summary>
 		/// Waits until the next frame
 		/// </summary>
 		public static readonly WaitUntil NextFrame = new WaitUntil(1, WaitType.Frames);
 
-		private readonly WaitType type;
 		private float internalValue;
+		private readonly Func<bool> subroutineMoveNext;
```
Barsonax, line 26: To generalize this more make this return a int and call it updaterfunction so we can use for other things as well.

Source/Core/Duality/Utility/Coroutines/WaitUntil.cs
```diff
@@ -13,16 +13,18 @@ private enum WaitType
 		{
 			Frames,
 			GameTime,
-			RealTime
+			RealTime,
+			Coroutine
```
Barsonax, line 17: Rename to Function

Source/Core/Duality/Utility/Coroutines/WaitUntil.cs
```diff
@@ -50,6 +60,10 @@ internal void Update()
 				case WaitType.RealTime:
 					this.internalValue -= Time.UnscaledDeltaTime;
 					break;
+
+				case WaitType.Coroutine:
+					this.internalValue = this.subroutineMoveNext() ? float.PositiveInfinity : 0;
```
Barsonax, line 65: Since the func will return a int now you can just assign the result to internalValue

Source/Core/Duality/Utility/Coroutines/WaitUntil.cs
```diff
@@ -84,5 +98,19 @@ public static WaitUntil TimeSpan(TimeSpan timeSpan, bool realTime = false)
 		{
 			return WaitUntil.Seconds((float)timeSpan.TotalSeconds, realTime);
 		}
+
+		/// <summary>
+		/// Waits until the
```
Barsonax, line 103: Sentence is not finished

Barsonax:
Scanned through the changes a bit. Looks pretty good already but I think we can generalize it a bit more so we can use this for more than just a coroutineends feature.

Also I want to measure the performance of this again.

Barsonax:
As expected some allocations are happening. Seems to be about 144 bytes per call which does impact performance quite a bit.

```
// * Summary *

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.836 (1909/November2018Update/19H2)
AMD Ryzen 9 3900X, 1 CPU, 24 logical and 12 physical cores
  [Host]     : .NET Framework 4.8 (4.8.4180.0), X64 RyuJIT
  DefaultJob : .NET Framework 4.8 (4.8.4180.0), X64 RyuJIT


|                Method |      N |             Mean |          Error |         StdDev |     Gen 0 |     Gen 1 |    Gen 2 |  Allocated |
|---------------------- |------- |-----------------:|---------------:|---------------:|----------:|----------:|---------:|-----------:|
| CallEndlessSubRoutine |      1 |         72.61 ns |       0.835 ns |       0.781 ns |    0.0869 |         - |        - |      144 B |
| CallEndlessSubRoutine |     10 |        672.92 ns |       3.946 ns |       3.498 ns |    0.8698 |         - |        - |     1444 B |
| CallEndlessSubRoutine |    100 |      6,785.61 ns |      32.611 ns |      27.232 ns |    8.6975 |         - |        - |    14443 B |
| CallEndlessSubRoutine |   1000 |     78,150.10 ns |     691.155 ns |     646.507 ns |   56.3965 |   14.0381 |        - |   144427 B |
| CallEndlessSubRoutine |  10000 |  1,027,735.81 ns |  11,053.970 ns |  10,339.890 ns |  230.4688 |  115.2344 |        - |  1444252 B |
| CallEndlessSubRoutine | 100000 | 29,019,037.50 ns | 248,125.998 ns | 232,097.212 ns | 2625.0000 | 1093.7500 | 343.7500 | 14442578 B |
```

Barsonax:
Not yet satisfied with the performance as there is alot of allocations going on (closures, boxing) but I don't think there's a way around this unless C# adds something like a `yield all` or `yield foreach` keyword.

In that regard writing this does exactly the same:
```cs
foreach(var step in SubCoroutine()) {
  yield return step;
}
```
This should prevent most of the allocations (there is still a enumerator being allocated) and makes the WaitUntil struct smaller.

It is however more verbose and less readable.

SirePi:
If this is how it is, then I believe it's better to just remove the method altogether.. there are many ways to deal with it in case.
You can take the foreach route, or you can go with the full coroutine-declaring approach that even lets you pause or resume it.

I'll just close it

PR #825, opened 2020-05-24
Added v4 template to the main repo

Barsonax:
Project generates a zip on build with the files from the template which can be used directly without other tools other than something to unpack it.

When you pack it it generates a nuget package that can be used with `dotnet new`. You can install it from nuget but you can also install it directly from a file path which is handy when testing:
```
dotnet new -i C:\git\duality\Source\DualityTemplates\bin\AdamsLair.Duality.Templates.4.0.0-alpha
```
Then create the folder for the solution, navigate to it and create the solution:
```
mkdir FlappyBird
cd FlappyBird
dotnet new dualitysolution --allow-scripts yes
```
This also builds the solution so you can use duality straight away (will probably fail if you do not have a localfeed setup with duality 4.0.0-alpha). It will get the name from the folder name but you can also specify it yourself with the `--name` parameter.

The nuget package is a template pack so it is possible to add multiple templates in the future (like a core plugin template, editor plugin template etc). Hence the naming is also generic.

To update the template just install it again with `dotnet new -i`

There is currently no integration with the nightly builder yet so the above all still to be done manually but thats fine for now as thats the same as the old situation I believe.

**NOTE**: template assumes the new folder structure in #827  

**Known issues**
- Name of the project you create with the template cannot contain spaces or dots. Not a very big issue but we might wanna look into solving this in a nice way for the future. In case you don't supply a name yourself it will take the name of the current folder so that folder name cannot have spaces or dots either then.


ilexp:
Since #823 was superseded by #827, is this template still up-to-date?

Barsonax:
No it has to be updated which I will when that other PR is merged

Barsonax:
@ilexp I updated the template with the new paths.

Barsonax:
@ilexp @SirePi Can anyone give this a review so this can be merged?

PR #827, opened 2020-05-29
revised project structure with dualityeditor still at the root

Barsonax:
Revision of #823

Decided to keep DualityEditor/DualityLauncher in the root for now as that doesn't require much changes in duality:
```
[Project Folder]
|
+- [Project].sln
|
+- DualityEditor.exe
+- Data
+- Plugins
+- all the other project files and binaries
|
+- Import
|
+- Source
   +- Launchers
      +- GameLauncher
      +- GameEditor
   +- Plugins
      +- GamePlugin
```


ilexp:
Is this still WiP? What's needed to complete this?

Barsonax:
Now its ready, also removed auto creating the source folder and added a check around the source folder filewatcher to prevent it from raising a error if the source folder does not exist.

I believe this is more inline with the role of source folder in v4 where duality no longer manages it but just integrates with it.

ilexp:
Looking good, approval still stands 👍

ilexp:
Can the feature branch be deleted, or do you still need it?

Issue #820, opened 2020-05-15
Duality v4.0 Release Schedule


ilexp:
### Summary

This is an aggregate issue to reflect what still needs to be done before [Duality v4.0](https://github.com/AdamsLair/duality/milestone/8) can be released, intended to roughly list points that did not make it into their own issues yet. Some of these points may already be done partially or fully and are listed to verify their current state. 

The intent of this is to answer and evolve the question "where are we on this?", and act as an internal roadmap and discussion platform for v4.0.

### Analysis

- Move the [DualityProjectTemplate](https://github.com/Barsonax/DualityProjectTemplate) that we'll bundle in the download .zip to somewhere inside this repo.
  - Consider adding an automated .zip pack step in some .csproj, like it was done for the now obsolete embedded [GamePluginTemplate](https://github.com/AdamsLair/duality/tree/04a64eed2fcd846de720917e77fa47facce04a4a/Source/Editor/DualityEditor/EmbeddedResources/GamePluginTemplate), so it doesn't need to be zipped manually for release.
  - Remove the now obsolete "package .zip" build step from NightlyBuilder.
- Do some extended testing with the template to make sure everything works and is reasonably polished.
  - Consider releasing a set of `4.0-beta` pre-release packages and a `4.0-beta` download .zip that can be used by developers and the community early adopters for testing prior to release.
- Update the [download link](https://get.duality2d.net/) subdomain to always point to the [latest release download](https://github.com/AdamsLair/duality/releases/latest/download/Duality.zip), see [here](https://help.github.com/en/github/administering-a-repository/linking-to-releases).
- Add a v4 docs folder including any required page updates and a new changelog page, see issue #799 for that.
- Consider to also do and bundle any other (breaking) changes that would require a major version step, preferably those that would benefit from the netstandard 2.0 / C# 7.3 upgrade.
  - We don't have an issue for that yet, but `RawList<T>` for example could use an API upgrade with a ref-return indexer, and so on.
  - Nothing specific required here, but it would be a good opportunity. Also, this should probably be in beta for a while, and if we're waiting for release anyway, why not do something useful with the time 😄 
- Eventually, when all is tested and done, bump all package versions to `4.0` and merge into `release`.
[...]

PR #830, opened 2020-05-30
Improved RawList API, Implementation and Tests

ilexp:
As noted in #820, the `RawList<T>` class can benefit from newly available ref-returns in C# 7.3 for its indexer implementation. Since this is a breaking change for v4.0, it also was a good opportunity to rework and extend some of its API to better address its core use case.

This PR adds several small improvements, as well as a by-ref indexer for `RawList<T>`. For a detailed list, see the individual commits, which have been structured so each one represents a group of changes.

### DIFF
--- a/Source/Core/Duality/Utility/RawList.cs
+++ b/Source/Core/Duality/Utility/RawList.cs
@@ -135,7 +135,9 @@ namespace Duality
 		/// <param name="item"></param>
 		public void Add(T item)
 		{
-			this.Insert(this.count, item);
+			this.Reserve(this.count + 1);
+			this.data[this.count] = item;
+			this.count++;
 		}
 		/// <summary>
 		/// Adds a range of new items to the list.
@@ -381,7 +383,7 @@ namespace Duality
 		public void Reserve(int capacity)
 		{
 			if (this.data.Length >= capacity) return;
-			Array.Resize(ref this.data, MathF.Max(this.data.Length * 2, capacity, BaseCapacity));
+			Array.Resize(ref this.data, Math.Max(Math.Max(this.data.Length * 2, capacity), BaseCapacity));
 		}
 		/// <summary>
 		/// Moves a range of elements by a certain value and resets the indices that now remain empty.
--- a/Source/Core/Duality/Utility/RawList.cs
+++ b/Source/Core/Duality/Utility/RawList.cs
@@ -65,21 +65,17 @@ namespace Duality
 		}
 
 		/// <summary>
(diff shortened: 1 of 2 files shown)


Source/Core/Duality/Utility/RawList.cs
```diff
@@ -65,49 +66,40 @@ public int Capacity
...
 
 		/// <summary>
-		/// [GET / SET] A safety-checked index accessor to the lists internal array. Will throw an <see cref="System.IndexOutOfRangeException"/>
+		/// [GET / SET] A safety-checked by-ref index accessor to the lists internal array. Will throw an <see cref="System.IndexOutOfRangeException"/>
 		/// when attempting to access indices exceeding <see cref="Count"/>.
 		/// </summary>
 		/// <param name="index"></param>
-		public T this[int index]
+		/// <returns></returns>
+		public ref T this[int index]
```
Barsonax, line 74: Put a `[DebuggerBrowsable(DebuggerBrowsableState.Never)]` here so its consistent with the explicit interface implementations

Barsonax:
For the performance improvements did you do any testing (benchmarkdotnet?)? Pretty sure this improves the performance but its always good to verify and also know how much it improves it.

Source/Core/Duality/Utility/RawList.cs
ilexp: Now that you mention it, this attribute doesn't make sense for any of the indexers. Not sure why I put them there, removed all of them. Thanks!

ilexp:
Addressed the change request you made, see above.

> For the performance improvements did you do any testing (benchmarkdotnet?)? Pretty sure this improves the performance but its always good to verify and also know how much it improves it.

You're right, but no, I didn't run any benchmarks. Mostly for time reasons and given the very small scope of the two tweaks, I stopped at "pretty sure it's an improvement".

Barsonax:
> Addressed the change request you made, see above.
> 
> > For the performance improvements did you do any testing (benchmarkdotnet?)? Pretty sure this improves the performance but its always good to verify and also know how much it improves it.
> 
> You're right, but no, I didn't run any benchmarks. Mostly for time reasons and given the very small scope of the two tweaks, I stopped at "pretty sure it's an improvement".

Ok well as the changes are pretty small and seem to make sense iam assuming it will be ok :)

Source/Core/Duality/Utility/RawList.cs
Barsonax: Haha, I thought you put them there on purpose :P

PR #840, opened 2020-06-15
Added a launcher class and use this to launch duality in the template

Barsonax:
Untested and mostly intended for discussion at of this moment.

Changes:
- Put the logic to run duality in a separate Run method.

Future:
- Expose some configuration so that this run method can be used in different contexts (such as unit tests).


ilexp:
No time for a propre review yet, but one thing I noticed were `NativeMethods` used by the launcher being moved into the `Duality` core. However, the core is supposed to be strictly platform-agnostic, and those launcher methods are Windows only - so they can't be moved into the core.

What they're dealing with is manually opening a console window in a non-console application, which is purely for development / debug purposes. I think it would be completely fine if the launcher API did not have this functionality at all and we'd leave it as a local add-on in the "official" launcher application. Custom launchers would then omit this feature unless users copy-pasted it.

Barsonax:
Moved the nativemethods part back

ilexp:
Can we separate the version bump from this PR? Adds a lot of noise to the review and it's also mixing two things that are probably better off independent. I mean, we might add any number of tweaks and fixes before releasing the next alpha, and a version bump commit should always be the final change for its version number.

Barsonax:
> Can we separate the version bump from this PR? Adds a lot of noise to the review and it's also mixing two things that are probably better off independent. I mean, we might add any number of tweaks and fixes before releasing the next alpha, and a version bump commit should be the final change for its version number.

Yeah I was thinking the same, just needed the version bump to test it (else it takes the one on nuget.org...)

Source/Launcher/DualityLauncher.csproj
```diff
@@ -8,24 +8,25 @@
...
-    <ProjectReference Include="..\Core\Primitives\DualityPrimitives.csproj">
-      <Project>{c87a6a2f-e537-48fa-a789-3c54cf29cc09}</Project>
-      <Name>DualityPrimitives</Name>
-    </ProjectReference>
-    <ProjectReference Include="..\Core\Duality\Duality.csproj">
-      <Project>{D821AFB9-BCC9-4025-9A9C-798E2BC546AB}</Project>
-      <Name>Duality</Name>
-    </ProjectReference>
-  </ItemGroup>
+	<ItemGroup>
```
Barsonax, line 16: Seems it now uses tabs for indenting even though the editor config should set it to spaces for csproj files..

Source/Core/Duality/Launcher/DualityLauncher.cs
```diff
@@ -0,0 +1,94 @@
...
+				DualityApp.Terminate();
+			}
+
+			// Clean up the log file
+			if (logfileWriter != null)
+			{
+				Logs.RemoveGlobalOutput(logfileOutput);
+				logfileWriter.Flush();
+				logfileWriter.Close();
+				logfileWriter = null;
```
SirePi, line 80: Could remove the null assignments, as the program is exiting and they no longer used anyway

Barsonax:
> Expected to see the old Launcher project converted to a library and have a one-line command in the Template's Main, but this way works as well.
> 
> Clicked wrong.. should be "Requested Changes"

Yeah but the launcher had some native code in it which cannot be in core and we also still need a executable for development.

ilexp:
So what this PR does, if I'm reading this correctly:

- Removes the DualityLauncher executable and package entirely.
- Adds Launcher API to the Duality core, which is essentially a convenience wrapper around functionality that was previously invoked manually in the launcher executable.
- Transforms the game launcher dummy project in the solution template into an actually compiled executable, which uses to launcher API to create a custom launcher for each game.
  - This will allow devs to easily specify custom icons, launcher names, or even change the entire startup process.
- The whole dev / debug console code that wouldn't be compatible with core restrictions remains in the solution template.

Sounds pretty good to me 👍 

Can we extend this one step further and make the `dotnet new` package auto-rename the launcher project to match the game? That way, a user-created `TestGame` project will have a `TestGame.exe` launcher right from the start.

We could even bundle a copy of the [default embedded `EditorUserData.xml`](https://github.com/AdamsLair/duality/blob/833e7ecd0af0268233671a84c0fc546544a2c9c7/Source/Editor/DualityEditor/EmbeddedResources/editoruserdata.xml) in the solution template, with an added [relative `LauncherPath` entry](https://github.com/AdamsLair/duality/blob/833e7ecd0af0268233671a84c0fc546544a2c9c7/Source/Editor/DualityEditor/DualityEditorApp.cs#L528) that is replaced by the `dotnet new` template logic, so the editor is correctly configured to use the custom launcher exe right away.

Barsonax:
The template already changes the exe name.

You're right that we do need a editorconfig with a modified path now though

ilexp:
> You're right that we do need a editorconfig with a modified path now though. However might be better to do that in a future PR as to not make this one too complicated (need to add a EditorAppData.xml etc)

Created a issue #848 for this 👍

Source/Launcher/Program.cs
```diff
@@ -1,13 +1,5 @@
 ﻿using System;
```
ilexp, line 1: Why do we still have a launcher in the core project on its own? Since there is now the convenience API in the core and the launcher code in the template, this should be obsolete?

Barsonax: We still need it to debug during developing duality ourselves. However we could just reference the same source files.

Barsonax: The same source code files as in the template are now being used

PR #842, opened 2020-06-15
fix dotnet build not working

Barsonax:
While adding a benchmarkdotnet project I noticed some problems when using the dotnet cli. This PR aims to fix these problems:
- SolutionDir is not set when running dotnet build on a project. Solved this by simply adding a RootFolder variable and use that instead of SolutionDir. Before this fix you would get a cryptic sourcelink error:
```
C:\Users\Rick\.nuget\packages\microsoft.sourcelink.common\1.0.0-beta2-19351-01\build\Microsoft.SourceLink.Common.targets(50,5): error : SourceRoot.SourceLinkUrl must contain a single wildcard '*': 'file:///raw/C:/git/duality/+refs/heads/*:refs/remotes/Barsonax/*/951bdcb2edba7800135da8abe7610a8b791a29e6/*'" [C:\git\duality\Source\Core\Primitives\DualityPrimitives.csproj]
```
- Updated sourcelink as it has been out of beta for a while. This also turned the above error into this which made it clear the issue was with SolutionDir being undefined:
```
C:\git\duality\Source\Core\Primitives\DualityPrimitives.csproj(11,5): error MSB3021: Unable to copy file "bin\Release\netstandard2.0\DualityPrimitives.dll" to "*Undefined*Build\Output\DualityPrimitives.dll". The filename, directory name, or volume label syntax is incorrect. : 'C:\git\duality\Source\Core\Primitives\*Undefined*Build\Output'
```
- Set GenerateResourceUsePreserializedResources to true to solve a difference between a build in visual studio and the dotnet cli. This seems to be only a issue when targetting netframework. https://github.com/microsoft/msbuild/issues/4704. On .net core this is true by default.
```
C:\Program Files\dotnet\sdk\3.1.300\Microsoft.Common.CurrentVersion.targets(3032,5): error MSB3823: Non-string resources require the property GenerateResourceUsePreserializedResources to be set to true. [C:\git\duality\Test\Core\DualityTests.csproj]
C:\Program Files\dotnet\sdk\3.1.300\Microsoft.Common.CurrentVersion.targets(3032,5): error MSB3822: Non-string resources require the System.Resources.Extensions assembly at runtime, but it was not found in this project's references. [C:\git\duality\Test\Core\DualityTests.csproj]
```

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -4,8 +4,10 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <NoWarn>1701;1702;CS1591;CS0419</NoWarn>
     <DefaultItemExcludes>$(DefaultItemExcludes);Content\**</DefaultItemExcludes>
+    <RootFolder>$(MSBuildThisFileDirectory)</RootFolder>
+    <GenerateResourceUsePreserializedResources>true</GenerateResourceUsePreserializedResources>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-19351-01" PrivateAssets="All"/>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
   </ItemGroup>
 </Project>
--- a/Source/Core/Duality/Duality.csproj
+++ b/Source/Core/Duality/Duality.csproj
@@ -82,10 +82,10 @@
 	</ItemGroup>
 
 	<Target Name="AggregateOutput" BeforeTargets="AfterBuild">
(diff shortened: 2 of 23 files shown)


ilexp:
Why is `SolutionDir` undefined though? Is it a deprecated? What's the difference between the `dotnet` CLI and regular `msbuild` here?

Barsonax:
> Why is `SolutionDir` undefined though? Is it a deprecated? What's the difference between the `dotnet` CLI and regular `msbuild` here?
> 
> > Updated sourcelink as it has been out of beta for a while.
> 
> 👍

Its only defined if you build the solution. If you build a project its undefined. When running benchmarks you want to be able to run a dotnet command directly on the project.

By putting it in a props file the variable is now always available.

ilexp:
Ah, makes sense.

PR #843, opened 2020-06-15
Migrate performance benchmarks in unit tests to benchmarkdotnet

Barsonax:
Moved all our benchmarks in our unit tests to a separate benchmarkdotnet project.

How to use:
- cd to Benchmarks\DualityBenchmarks
- Run `dotnet run -c Release --list tree` to view the available benchmarks
- Run `dotnet run -c Release --filter CloneProviderBenchmarks` to run the CloneProviderBenchmarks

TODO:
- Some benchmarks are running alot of itterations. This is a leftover of the setup in the unit tests and is usually not needed unless you are interested in how something scales. Change this to only run once or if scaling is important parameterize it so its easier to see the scaling (for example do 3 runs with the first run only 1 itteration, the second one 10 itterations etc). DONE

Rebased on https://github.com/AdamsLair/duality/pull/842

### DIFF
--- a/Benchmarks/DualityBenchmarks/CloneProviderBenchmarks.cs
+++ b/Benchmarks/DualityBenchmarks/CloneProviderBenchmarks.cs
@@ -9,33 +9,25 @@ namespace DualityBenchmarks
 	public class CloneProviderBenchmarks
 	{
 		private Random rnd;
-		private TestObject[] results;
 		private TestObject data;
 
 		[GlobalSetup]
 		public void Setup()
 		{
 			this.rnd = new Random(0);
-			this.results = new TestObject[200];
 			this.data = new TestObject(this.rnd, 5);
 		}
 
 		[Benchmark]
-		public void CloneTestObjectGraph()
+		public TestObject CloneTestObjectGraph()
 		{
-			for (int i = 0; i < this.results.Length; i++)
-			{
-				this.results[i] = this.data.DeepClone();
-			}
+			return this.data.DeepClone();
 		}
 
 		[Benchmark(Baseline = true)]
-		public void CreateWithoutClone()
+		public TestObject CreateWithoutClone()
 		{
-			for (int i = 0; i < this.results.Length; i++)
-			{
-				this.results[i] = new TestObject(this.rnd, 5);
-			}
+			return new TestObject(this.rnd, 5);
 		}
 	}
 }
--- a/Benchmarks/DualityBenchmarks/CloneProviderBenchmarks.cs
+++ b/Benchmarks/DualityBenchmarks/CloneProviderBenchmarks.cs
@@ -5,6 +5,7 @@ using Duality.Tests.Serialization;
 
 namespace DualityBenchmarks
 {
+	[MemoryDiagnoser]
 	public class CloneProviderBenchmarks
 	{
 		private Random rnd;
--- a/Benchmarks/DualityBenchmarks/CloneProviderBenchmarks.cs
+++ b/Benchmarks/DualityBenchmarks/CloneProviderBenchmarks.cs
@@ -1,0 +1,34 @@
+﻿using System;
+using BenchmarkDotNet.Attributes;
+using Duality.Cloning;
+using Duality.Tests.Serialization;
+
+namespace DualityBenchmarks
+{
+	public class CloneProviderBenchmarks
+	{
+		[Benchmark]
+		public void CloneTestObjectGraph()
+		{
+			Random rnd = new Random(0);
+			TestObject data = new TestObject(rnd, 5);
+			TestObject[] results = new TestObject[200];
+
+			for (int i = 0; i < results.Length; i++)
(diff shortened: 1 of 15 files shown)


ilexp:
Should probably not merge this until #842 is addressed due to the rebase, reviews should be easier afterwards as well.

Barsonax:
As #842 has been merged this can now be reviewed.

Before putting all current performance test in the new project iam curious what do you think about this new setup? @ilexp @SirePi

ilexp:
Overall setup looks pretty good 👍 

However, this doesn't seem right:

devenv_2qweZIQbRw

Looks like a missing ignore condition in the `.csproj`?

Also, I'd consider copy-pasting `TestObject` next to the benchmark to avoid the dependency to the `DualityTests` project, since I'm not sure if that should be there. Not a change request though, just food for thought.

Barsonax:
> Overall setup looks pretty good 👍 
> 
> However, this doesn't seem right:
> 
> devenv_2qweZIQbRw
> 
> Looks like a missing ignore condition in the `.csproj`?

Probably, haven't spend much attention to polishing everything as I was mostly interested in how you find the benchmarkdotnet way of doing things.

> Also, I'd consider copy-pasting `TestObject` next to the benchmark to avoid the dependency to the `DualityTests` project, since I'm not sure if that should be there. Not a change request though, just food for thought.

Ye when we move all the benchmarks it should make it more clear we don't need the dependency. Though in benchmarks you might wanna reuse the same models as in the tests (atleast I did in Singularity so much that I moved those to a separate project even).

Benchmarks/DualityBenchmarks/CloneProviderBenchmarks.cs
```diff
@@ -0,0 +1,41 @@
...
+				this.results[i] = this.data.DeepClone();
+			}
+		}
+
+		[Benchmark(Baseline = true)]
+		public void CreateWithoutClone()
+		{
+			for (int i = 0; i < this.results.Length; i++)
+			{
+				this.results[i] = new TestObject(this.rnd, 5);
```
Barsonax, line 37: Some thing I noticed that here we are actually measuring the performance of the TestObject constructor (which is quite inefficient due to alot of unneeded allocations) which might not be the best baseline case but fixing that is out of the scope of this PR.

Source/Core/Duality/Serialization/Serializer.cs
```diff
@@ -317,10 +317,16 @@ protected virtual void OnDisposed(bool manually)
...
 			if (!this.CanRead) throw new InvalidOperationException("Can't read object from a write-only serializer!");
-			this.BeginReadOperation();
-			object result = this.ReadObjectData();
-			this.EndReadOperation();
-			return result;
+
+			try
+			{
+				this.BeginReadOperation();
+				return this.ReadObjectData();
```
Barsonax, line 324: When a exception occurred here the serializer would never call EndReadOperation. This in turn caused another exception when the serializer was disposed which would hide the real error. Might wanna move this fix to a separate PR or just leave it in here as thats easier. I put it in a separate commit just in case.

Source/Core/Duality/Serialization/Serializer.cs
```diff
@@ -349,9 +355,16 @@ public void ReadObject<T>(out T obj)
...
 		{
 			if (!this.CanWrite) throw new InvalidOperationException("Can't write object to a read-only serializer!");
-			this.BeginWriteOperation();
-			this.WriteObjectData(obj);
-			this.EndWriteOperation();
+
+			try
+			{
+				this.BeginWriteOperation();
+				this.WriteObjectData(obj);
```
Barsonax, line 362: put the try finally here as well to make it consistent with the read part

Benchmarks/DualityBenchmarks/CloneProviderBenchmarks.cs
ilexp: Good catch 👍 I was surprised by that result as well. 

(Agree that this is out of scope for this PR though.)

Source/Core/Duality/Serialization/Serializer.cs
ilexp: I'd be fine with leaving it in this PR, since it's in a separate commit 👍 Just have to remember not to squash-merge this one 🙂

Source/Core/Duality/Serialization/Serializer.cs
ilexp: 👍

Benchmarks/DualityBenchmarks/CloneProviderBenchmarks.cs
Barsonax: Yeah, benchmarkdotnet is just way more precise than anything you could probably hand write in a decent amount of time. Combined with the detailed allocation data that just makes it easier to spot these kind of things which is a good thing.

PR #844, opened 2020-06-15
Fixed VS Debug / Run for launchers in Duality project template

ilexp:
## Summary

In the current alpha2 release, the run / debug buttons for `GameLauncher` and `GameEditor` in Visual Studio are broken, as their working directory is off by one - probably a regression due to recent folder structure changes.

This PR fixes the launch settings and also provides more descriptive names for the "Run" button / debug profile.


Barsonax:
Good catch

PR #849, opened 2020-06-24
Cleanup launcher

Barsonax:


Source/Core/Duality/Launcher/DualityLauncher.cs
```diff
@@ -1,22 +1,17 @@
...
 {
 	public class DualityLauncher
 	{
-		public static void Run(string[] args)
+		public static void Run(LauncherArgs launcherArgs)
 		{
 			System.Threading.Thread.CurrentThread.CurrentCulture = System.Globalization.CultureInfo.InvariantCulture;
 			System.Threading.Thread.CurrentThread.CurrentUICulture = System.Globalization.CultureInfo.InvariantCulture;
 
-			bool isDebugging = System.Diagnostics.Debugger.IsAttached || args.Contains(DualityApp.CmdArgDebug);
```
Barsonax, line 16: Might be better to only set isDebugging to true if the actual parameter is passed.

Atm you will get different behavior when attaching the debugger depending if it happened before this line of code was executed or after. Better to just start the launcher with the debugging flag.

ilexp: Good idea, should make it more transparent how this behaves 👍

Source/Core/Duality/Launcher/DualityLauncher.cs
```diff
@@ -41,23 +43,23 @@ public static void Run(string[] args)
...
+				this.launcherArgs);
 
 			// Open up a new window
 			WindowOptions options = new WindowOptions
 			{
 				Size = DualityApp.UserData.WindowSize,
-				ScreenMode = isDebugging ? ScreenMode.Window : DualityApp.UserData.WindowMode,
-				RefreshMode = (isDebugging || isProfiling) ? RefreshMode.NoSync : DualityApp.UserData.WindowRefreshMode,
+				ScreenMode = this.launcherArgs.IsDebugging ? ScreenMode.Window : DualityApp.UserData.WindowMode,
+				RefreshMode = (this.launcherArgs.IsDebugging || this.launcherArgs.IsProfiling) ? RefreshMode.NoSync : DualityApp.UserData.WindowRefreshMode,
```
ilexp, line 60: This is unrelated to your changes, but since you're at it anyway, can you remove the effect `IsDebugging` has on `RefreshMode`? 

I'm no longer sure why I added this way back, but I think it's a bad idea to always debug with infinite FPS when the game actually runs _with_ VSync in production. Stuff just behaves slightly different at 60 FPS vs 500 FPS with variable frame time and `TimeMult` / `DeltaTime` isn't designed to deal with differences _that_ large. Better to let people optimize for / test in the case that players will actually experience.

Source/Core/Duality/Launcher/LauncherArgs.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Linq;
+
+namespace Duality.Launcher
+{
+	public class LauncherArgs
```
ilexp, line 6: As per best practices, please do XML docs for all public members - bonus points for the private ones 🙂

Source/Core/Duality/Launcher/DualityLauncher.cs
```diff
@@ -1,22 +1,24 @@
 ﻿using System;
 using System.IO;
-using System.Linq;
 using Duality.Backend;
 using Duality.Resources;
 
 namespace Duality.Launcher
 {
-	public class DualityLauncher
+	public class DualityLauncher : IDisposable
```
ilexp, line 8: We missed that in the previous PR, but can you add XML docs for all public members?

Source/DualityTemplates/Templates/SolutionTemplate/Source/Launchers/DualityGame/Program.cs
```diff
@@ -10,10 +8,14 @@ internal static class Program
 		[STAThread]
 		public static void Main(string[] args)
 		{
-			bool isDebugging = System.Diagnostics.Debugger.IsAttached || args.Contains(DualityApp.CmdArgDebug);
-			bool isRunFromEditor = args.Contains(DualityApp.CmdArgEditor);
-			if (isDebugging || isRunFromEditor) ShowConsole();
-			DualityLauncher.Run(args);
+			var launcherArgs = new LauncherArgs(args);
```
ilexp, line 11: Please avoid `var` where possible to match the current style guide.

Source/DualityTemplates/Templates/SolutionTemplate/Source/Launchers/DualityGame/Program.cs
```diff
@@ -10,10 +8,14 @@ internal static class Program
...
+
+			if (launcherArgs.IsDebugging|| launcherArgs.IsRunFromEditor) ShowConsole();
+
+			using (var launcher = new DualityLauncher(launcherArgs))
```
ilexp, line 15: See above 🙂

Source/Core/Duality/Launcher/DualityLauncher.cs
Barsonax: How about the ScreenMode as that one also changes when debugging?

Barsonax:
Changed the code a bit. Initializing duality now happens in the constructor. Actually starting the game loop now happens in the `Run` method.

Separating these makes it possible to use the launcher class also in the unit tests which reduces code duplication but more importantly makes sure its being tested often.

Source/DualityTemplates/Templates/SolutionTemplate/Source/Launchers/DualityGame/Program.cs
Barsonax: I think the style could be a bit better on this. When the type is obvious its not needed to repeat the type again. Thats just extra noise. In the above code its clear that the type is `LauncherArgs` as we just used the constructor.

The editor config actually even has a option for this. We just need to change `csharp_style_var_when_type_is_apparent = false:suggestion` to `csharp_style_var_when_type_is_apparent = true:suggestion`.

ilexp:
Moving this discussion out of the code, since it's more general.

> I think the style could be a bit better on this. When the type is obvious its not needed to repeat the type again. Thats just extra noise (also we ship this to our users so we should make it as clean as possible). In the above code its clear that the type is `LauncherArgs` as we just used the constructor.
> 
> The editor config actually even has a option for this. We just need to change `csharp_style_var_when_type_is_apparent = false:suggestion` to `csharp_style_var_when_type_is_apparent = true:suggestion`.

First off, I won't reject your PR if you insist on using var in those two specific spots, because I think you've good reason to choose that way. I'd prefer it the other way around for consistency reasons, which you'll probably understand too, so I'd just leave it to you to reconcile the two arguments and priorities like you see fit.

----

Now to the general style guide thing and editor config change:

You're right that it doesn't add more information when you immediately populate a variable with a newly constructed value, but it does in all other cases. Having a style guide special case here, in my opinion, only leads to confusion as to where to use or not to use `var`. The current baseline of "just avoid `var` by default" is easy to remember and apply.

Editor / IDE hints aside, if we adopt a guideline based on "when the type is apparent", I fear that we may be entering a slippery slope. You're talking about this one here:

```csharp
var args = new LauncherArgs();
```

But then there's also:

```csharp
var args = this.GetLauncherArgs();
var args = this.launcherArgs;
var fooValue = 42.0f;
var fooValue = 42.0;
```

"Just avoid `var`" is dead easy to do and a clear guideline with no mental overhead, which is why I'd keep it as-is.

And honestly, I don't think this one is so bad:

```csharp
LauncherArgs args = new LauncherArgs();
```

Sure, that one word is there twice. So what? 😄 Code is very structured, and sometimes that structure involves repetition. Us developers are used to that anyway. Personally, I think it even makes it easier to read.

Barsonax:
> Moving this discussion out of the code, since it's more general.
> 
> > I think the style could be a bit better on this. When the type is obvious its not needed to repeat the type again. Thats just extra noise (also we ship this to our users so we should make it as clean as possible). In the above code its clear that the type is `LauncherArgs` as we just used the constructor.
> > The editor config actually even has a option for this. We just need to change `csharp_style_var_when_type_is_apparent = false:suggestion` to `csharp_style_var_when_type_is_apparent = true:suggestion`.
> 
> First off, I won't reject your PR if you insist on using var in those two specific spots, because I think you've good reason to choose that way. I'd prefer it the other way around for consistency reasons, which you'll probably understand too, so I'd just leave it to you to reconcile the two arguments and priorities like you see fit.

I know, just running into this at that specific line of code :P. 

> Now to the general style guide thing and editor config change:
> 
> You're right that it doesn't add more information when you immediately populate a variable with a newly constructed value, but it does in all other cases. Having a style guide special case here, in my opinion, only leads to confusion as to where to use or not to use `var`. The current baseline of "just avoid `var` by default" is easy to remember and apply.
> 
> Editor / IDE hints aside, if we adopt a guideline based on "when the type is apparent", I fear that we may be entering a slippery slope. You're talking about this one here:
> 
> ```cs
> var args = new LauncherArgs();
> ```
> 
> But then there's also:
> 
> ```cs
> var args = this.GetLauncherArgs();
> var args = this.launcherArgs;
> var fooValue = 42.0f;
> var fooValue = 42.0;
> ```

Actually in all these cases the IDE will suggest you to use a explicit type even with that option turned on as the type is not apparent. Methods, fields and properties could return anything regardless of their name. 42.0 is a double in C# and 42.0f is a float in C# but thats only clear if you know C#.

> "Just avoid `var`" is dead easy to do and a clear guideline with no mental overhead, which is why I'd keep it as-is.
> 
> And honestly, I don't think this one is so bad:
> 
> ```cs
> LauncherArgs args = new LauncherArgs();
> ```
> 
> Sure, that one word is there twice. So what? 😄 Code is very structured, and sometimes that structure involves repetition. Us developers are used to that anyway. Personally, I think it even makes it easier to read.

I agree its a small change but it does make the code more concise and its not that hard to change the editorconfig style.

However since this PR is not about changing the style iam gonna leave that out of this PR and move it to this issue https://github.com/AdamsLair/duality/issues/854. There might be some other style changes we would like to do as well.

PR #852, opened 2020-06-24
Add project settings and move launcher path to it

Barsonax:


ilexp:
> Ok for me, just a little thing bugging my brain: we have UserData and AppData, so why not ProjectData? 😄

To me, ProjectSettings just triggers the right associations as to what one should expect to find inside, and it's a term that's well known from other software and Unity. It's also too different from `UserData` and `AppData`, which are both read or in case of user data even written by the game while playing, when `ProjectData` is really just a development time thing.

Barsonax:
@ilexp when publishing the game editor files wouldn't be included right?

We have a number of different settings files:

For the game:
- `AppData.dat`
- `UserData.dat`

For the editor:
- `EditorUserData.xml` (uses some custom code flow to write to the file)
- `ProjectSettings.dat` (newly introduced in this PR)
- `DesignTimeData.dat` (not sure whats in here, its not in a readable format)

And some default variants of these files. We might wanna look into cleaning this up so we treat all these settings in the same way but thats for a different issue https://github.com/AdamsLair/duality/issues/855.

I agree with sirepi that the naming for `ProjectSettings.dat` might not be optimal. Personally I think since we already have `AppData.dat` for the game simply calling it `EditorAppData.dat` makes sense as it has the same purpose but for the editor.

ilexp:
> @ilexp when publishing the game editor files wouldn't be included right?

It depends - it's perfectly valid to deploy the editor with the game to give players a map editor or something. In that case, the project settings / (non-user) editor settings would be included as well. Perhaps cleaned up in some form, but that's a separate issue for later I think.

> I agree with sirepi that the naming for `ProjectSettings.dat` might not be optimal. Personally I think since we already have `AppData.dat` for the game simply calling it `EditorAppData.dat` makes sense as it has the same purpose but for the editor.

I still think `EditorAppData` is rather opaque as a term, not giving you a good idea on its contents, whereas `ProjectSettings` already gives you a rough direction. I'd rather clean up the naming of `AppData` and `UserData` into something more descriptive tbh. `UserData` should, judging from its contents, rather be a name with something like game options, game settings, or similar. And `AppData` might be better off with something like system config, engine config, app info. Having all the names in one consistent pattern is neat and tidy, but it doesn't help if nobody except us gets it without studying the docs.

_Edit:_ Maybe let's move this discussion over to #855? I've commented there with some more thoughts on this.

> `DesignTimeData.dat`

It's non-global / per-object data that is only relevant in the editor, [see here](https://github.com/AdamsLair/duality/blob/eb81dc7263588319c6603a651e508b06b6ad9ee1/Source/Editor/DualityEditor/DesignTimeObjectData.cs). Right now it only stores whether an object is locked or hidden. Could probably be refactored to be merged into any of the other files, depending on how we define the scope of this info, or stay separate for a while until we happen across a good idea on what else to do with it.

Barsonax:
For now I will keep the name as is in this PR. Lets change the name to the final name after discussing it in the issue.

PR #857, opened 2020-06-30
Added a duality test launcher for use in unit tests/benchmarks etc

Barsonax:
After making debugging benchmarks a bit easier and more strict I found out that we actually get a exception in the cleanup of the serialization benchmark.

Changes
- Adds a DualityTestLauncher which purpose is to launch duality in such a way that it works well with unit tests, benchmarks etc. I used this in the SerializerBenchmarks so that it uses the same environment that the unit tests are using.
- Added a warning that when a benchmark is run and a debugger is attached that DebugInProcessConfig is used since this might affect benchmarks results.
- Added ExecutionValidator.FailOnError to the benchmark runner when debugger is attached to make it easier to find errors. What this does is run each benchmark only 1 itteration just to verify if its working or not before doing the actual benchmark.

We might want to consider putting the DualityTestLauncher class in core or putting it in a separate nuget package as users might wanna use this class for their own unit tests.

EDIT: seems changing the working directory was to fix something nunit specific so having a separate DualityTestLauncher is not needed

### DIFF
--- a/Benchmarks/DualityBenchmarks/DualityBenchmarks.csproj
+++ b/Benchmarks/DualityBenchmarks/DualityBenchmarks.csproj
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net472</TargetFramework>
+    <OutputPath>$(RootFolder)Build\Output\</OutputPath>
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Benchmarks/DualityBenchmarks/Program.cs
+++ b/Benchmarks/DualityBenchmarks/Program.cs
@@ -2,6 +2,7 @@
 using System.Diagnostics;
 using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Validators;
 
 namespace DualityBenchmarks
 {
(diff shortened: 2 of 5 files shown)


Barsonax:
Hmm just noticed that the working directory only changes to `C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE` when running the unit tests.

In the benchmarks its already correct so no need actually to switch the working directory

ilexp:
> Or modify LoadPlugin to do this for you. @ilexp

Actually, I don't think `Assembly.Location` is defined for assemblies that were anonymously loaded from a `byte[]`, as is done for all plugins to avoid various issues. And since the overload you mentioned is used by both the regular load and reload methods, the path needs to be passed. It's just the one special case for unit test classes that's different.

Barsonax:
> > Or modify LoadPlugin to do this for you. @ilexp
> 
> Actually, I don't think `Assembly.Location` is defined for assemblies that were anonymously loaded from a `byte[]`, as is done for all plugins to avoid various issues. And since the overload you mentioned is used by both the regular load and reload methods, the path needs to be passed. It's just the one special case for unit test classes that's different.

Correct. Location is null I think or throws a error when the assembly is loaded from a byte array.

> ```cs
> 			// Manually register pseudo-plugin for the Unit Testing Assembly
> 			this.unitTestPlugin = DualityApp.PluginManager.LoadPlugin(
> 				typeof(DualityTestsPlugin).Assembly,
> 				typeof(DualityTestsPlugin).Assembly.Location);
> ```
> 
> This should be fine though 👍 

Will make a PR to cleanup that part later.

PR #858, opened 2020-07-01
Cleanup InitDualityAttribute's

Barsonax:
Just a small cleanup since nunit provides a utility for setting the current directory already.

Also use `typeof(DualityTestsPlugin).Assembly.Location` for the location of the test plugin as thats cleaner and also actually fixes a potential bug as previously it would be set to `Duality.dll` instead of `DualityTests.dll`.

### DIFF
--- a/Test/Core/InitDualityAttribute.cs
+++ b/Test/Core/InitDualityAttribute.cs
@@ -1,8 +1,9 @@
 ﻿using System;
-using System.IO;
+
+using Duality.Launcher;
+
 using NUnit.Framework;
 using NUnit.Framework.Interfaces;
-using Duality.Launcher;
 
 namespace Duality.Tests
 {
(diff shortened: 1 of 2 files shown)


Test/Core/InitDualityAttribute.cs
```diff
@@ -25,21 +26,18 @@ public void BeforeTest(ITest details)
 
 			// Set environment directory to Duality binary directory
 			this.oldEnvDir = Environment.CurrentDirectory;
-			string codeBaseURI = typeof(DualityApp).Assembly.CodeBase;
-			string codeBasePath = codeBaseURI.StartsWith("file:") ? codeBaseURI.Remove(0, "file:".Length) : codeBaseURI;
-			codeBasePath = codeBasePath.TrimStart('/');
-			Console.WriteLine("Testing Core Assembly: {0}", codeBasePath);
-			Environment.CurrentDirectory = Path.GetDirectoryName(codeBasePath);
+			Console.WriteLine("Testing Core Assembly: {0}", TestContext.CurrentContext.TestDirectory);
+			Environment.CurrentDirectory = TestContext.CurrentContext.TestDirectory;
```
ilexp, line 30: The new code is only equivalent to the old one when the testing binary is in the same folder as `Duality.dll`, which formerly was `Build/Output` in both cases. I don't currently have the full overview with all the recent changes, but if this is still true this should work out.

ilexp: However, the console log is no longer correct: The printed path is actually the directory, not the assembly (which is fine, but not as advertised). Same for the console log in the editor part down below. Maybe just change the line to something like "Testing in working directory: '{0}'"?

PR #860, opened 2020-07-03
Use duality serializers for EditorUserData

Barsonax:
Changes:
- Added a new way to save editoruserdata using duality serializers. Just serialize and deserialize models instead of XElements
- Old system still works, it gets stored in a string field in PluginSettings. This prevents us from having to rewrite everything that uses this at the same time. Default settings have moved to their respective projects though with the use of `EditorPlugin.GetDefaultUserData`.
- Dockpanel xml is stored in a separate string field on the DualityEditorUserData class

Issues:
- Camview editor plugin has not been converted to the new api yet so for now keeping both systems intact. Refactorings where done in separate commits to make them easier to review or even split into a separate PR.

TODO
- [ ] xml comments


Barsonax:
rebase on master after #856 has been merged and then update it with the new changes introduced in that PR

ilexp:
Superseded by #865 

@Barsonax Please document the relation of duplicate PRs / issues when closing them - makes it a lot easier to read when tracing back progress later on 🙂

Barsonax:
In case anyone in the future reads this. This PR was closed as I splitted up the changes into multiple PR's to make reviewing easier.

PR #861, opened 2020-07-05
Added XElement support to the serializer

Barsonax:
Serializing and deserializing `XElement` did not work properly before. This PR adds a `XElementSurrogate` which allows all serializers in duality to properly handle `XElement`.

Additionally `XmlSerializer` has a small extra modification to make the serialized xml more readable, without this change the serialized xml would end up as a base64 string.

TODO
- [x]  unittests
- [x] comments

### DIFF
--- a/Source/Core/Duality/Serialization/Surrogates/XElementSurrogate.cs
+++ b/Source/Core/Duality/Serialization/Surrogates/XElementSurrogate.cs
@@ -1,0 +1,26 @@
+﻿using System.Xml.Linq;
+
+namespace Duality.Serialization.Surrogates
+{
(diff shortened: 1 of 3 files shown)


Source/Core/Duality/Serialization/Surrogates/XElementSurrogate.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Xml.Linq;
+
+namespace Duality.Serialization.Surrogates
+{
+	public class XElementSurrogate : SerializeSurrogate<XElement>
```
ilexp, line 5: XML comments on this class and members please 🙂

ilexp:
Test comment to check webhook

PR #867, opened 2020-07-13
update camview settings

Barsonax:


Barsonax:
Gonna merge this so that camview works properly again even though its still using xml based editoruserdata (with the change being that those settings now reside in the plugin). Lets refactor this to use a model in a future PR.

Issue #855, opened 2020-06-27
Refactor/rethink how duality handles settings


Barsonax:
### Summary

There are some things that might be improved in the settings department of duality:

First it seems that duality uses 2 approaches to storing settings.
- `EditorUserData.xml` seems to use a completely separate code flow for writing to the settings file in `DualityEditorApp.SaveUserData()`. The file itself also has 2 `<?xml version="1.0" encoding="utf-8"?>` which I have never seen before in a single file. Maybe some hack around dockpanel behavior?
- Other setting files are using the duality `Serializer` class to write to the files. They also use the `.dat` file extension instead of `.xml`

Second there's not a clear location for where saving and loading is implemented. Sometimes it happens in `DualityApp` or `DualityEditorApp` and other times it happens in the settings class itself. I think we might benefit from making a interface with Save and Load methods where these are implemented on the class itself as that would allow us to treat settings in a more generic way (such as here https://github.com/AdamsLair/duality/blob/master/Source/Editor/DualityEditor/DualityEditorApp.cs#L1012) and also clean up some code from the already big  `DualityApp` and `DualityEditorApp` classes.

Thirdly because of how the serializer api works atm it always creates a new instance. This makes it impossible to put change events on the settings class itself because then you would also have to resubscribe to those events which kinda defeats the purpose. It might be beneficial if the serializer api could take a already existing instance and update that instance.

ilexp:
> Thirdly because of how the serializer api works atm it always creates a new instance. This makes it impossible to put change events on the settings class itself because then you would also have to resubscribe to those events which kinda defeats the purpose. It might be beneficial if the serializer api could take a already existing instance and update that instance.

Yeah, this is a design flaw in the serializer API, but a big one to fix properly. We should skip this for now and just assume the serializer API to be as-is for now. We should track this it in a separate issue though, because it is worth improving here.

> Second there's not a clear location for where saving and loading is implemented. Sometimes it happens in `DualityApp` or `DualityEditorApp` and other times it happens in the settings class itself. I think we might benefit from making a interface with Save and Load methods where these are implemented on the class itself as that would allow us to treat settings in a more generic way (such as here https://github.com/AdamsLair/duality/blob/master/Source/Editor/DualityEditor/DualityEditorApp.cs#L1012) and also clean up some code from the already big `DualityApp` and `DualityEditorApp` classes.

[...]

PR #868, opened 2020-07-13
updated logviewsettings

Barsonax:


ilexp:
## Progress

- Merged that latest changes from #865 and subsequent tweaks from master into this branch.
- Prepared a first draft on how the new settings object for the LogView plugin could be used, see [this comment](https://github.com/AdamsLair/duality/issues/855#issuecomment-662366759) for background info.

## ToDo

- Discuss and tweak the draft. If we're happy with it, proceed:
- Add XML docs for `LogViewSettings` class and members.
- Add XML docs to all added API in `LogView`.
- (Squash-)Merge PR and applying the same principle when addressing the others linked in #855.

ilexp:
Ping @Barsonax, since I can't request a review from you.

Source/Plugins/EditorModules/LogView/Modules/LogView.cs
```diff
@@ -1,26 +1,34 @@
...
+		public LogViewSettings UserSettings
+		{
+			get { return this.userSettings; }
+			set
+			{
+				if (this.userSettings != value)
+				{
+					this.userSettings = value;
+					this.ApplyUserSettings();
+				}
```
ilexp, line 27: Spotted a slight issue: Calling `Apply` only with a different instance will break when

- The `LoadUserData` code when it happens as part of a `SettingsContainer.Apply` call in the editor, since it will be the same instance, maybe only with changed field values and also when
- Settings serialization is one day extended to allow updating an existing instance.

So we should either call `ApplyUserSettings` explicitly after assignment in `LoadUserData`, or always apply on assignment. Currently leaning towards the first option, but not sure yet.

Barsonax:
Looks good. One thing is that I see you merged master into this branch. Its cleaner to rebase this branch on master as that prevents the merge commit from being generated. Rebasing simply puts the new commits of this branch on top of master so you end up with a cleaner history.

Source/Plugins/EditorModules/LogView/Modules/LogView.cs
Barsonax: > The LoadUserData code when it happens as part of a SettingsContainer.Apply call in the editor, since it will be the same instance, maybe only with changed field values and also when

In this case the developer has to call Apply himself. Not really a way around that I think.

> Settings serialization is one day extended to allow updating an existing instance.

If this happens then we don't really need a Settingscontainer anymore as we can just use the setting class directly. That setting class can then implement changed/changing events so we could even make the change events more fine grained.

Barsonax: Is this really a issue atm though? Before this refactor I don't think we even had a changed event for these pluginsettings do it not a regression.

In the future when we can just reuse the same instance we might look into implementing `INotifyPropertyChanged` or similar.

ilexp: > In this case the developer has to call Apply himself. Not really a way around that I think.

Yes, but the issue is that even when calling `Apply` on the global editor settings container, it will be ignored by the log view UI. I think having to call it is fine as (like you say) there's not really a great way around that right now - but the second part is just a bug.

It's not a regression, but since we're establishing a new model here, we should at least make sure to remain consistent within that model. We just didn't have this problem before because there were no settings exposed in code whatsoever.

It's nothing major to fix, probably just a few lines after choosing which way to go, so might as well do it 😁 I can do a quick followup commit as soon as I get around to it.

Barsonax:
Can't approve since its my own PR so gonna to that by comment: Approved.

Merging this.

PR #870, opened 2020-07-13
update sceneViewsettings

Barsonax:


Barsonax:
Ready for a review @ilexp @SirePi

Source/Plugins/EditorModules/SceneView/SceneViewSettings.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Duality.Editor.Plugins.SceneView
+{
+	public class SceneViewSettings
+	{
+		private bool showComponents = true;
+
+		public bool ShowComponents
```
ilexp, line 7: Class and property need an XML comment. (The property controls whether components are shown as individual nodes in the scene tree.)

Source/Plugins/EditorModules/SceneView/Modules/SceneView.cs
```diff
@@ -77,6 +74,14 @@ private class CreateContextEntryTag
...
 		private MenuModelItem	nodeContextItemPaste	= null;
 
+		private SceneViewSettings userSettings = new SceneViewSettings();
+
+		public SceneViewSettings UserSettings
+		{
+			get { return this.userSettings; }
+			set { this.userSettings = value; }
+		}
+
```
ilexp, line 84: Style consistency: Remove the newline between properties to group them closely.

Source/Plugins/EditorModules/SceneView/Modules/SceneView.cs
```diff
@@ -77,6 +74,14 @@ private class CreateContextEntryTag
 		private MenuModelItem	nodeContextItemCopy		= null;
 		private MenuModelItem	nodeContextItemPaste	= null;
 
+		private SceneViewSettings userSettings = new SceneViewSettings();
```
ilexp, line 77: Style consistency: Use a double newline to separate members of different kinds if either of them is grouped into semantic blocks on their own.

(This was probably wrong before your change, but might as well address it while here)

Source/Plugins/EditorModules/SceneView/Modules/SceneView.cs
```diff
@@ -171,17 +176,10 @@ protected override void OnClosed(EventArgs e)
 			Scene.ComponentAdded -= this.Scene_ComponentAdded;
 			Scene.ComponentRemoving -= this.Scene_ComponentRemoving;
 		}
-
-		internal void SaveUserData(XElement node)
-		{
-			node.SetElementValue("ShowComponents", this.buttonShowComponents.Checked);
-		}
-		internal void LoadUserData(XElement node)
+
```
ilexp, line 179: Style consistency: Another drive-by fix, this one should be a double newline as well.

PR #876, opened 2020-08-23
Fix display of version info to no longer include launcher version.

TheRut:
Found this bug while exploring the codebase, which was preventing the About dialog from displaying.


Barsonax:
Thanks for making the PR. 

I was able to reproduce this and found that I broke this in 5e484144224376f69aa1754ae728495d6ea76e28. I don't think we should put a try catch around this though but rather just change `labelVersionData.text` to be `"{0}\r\n{1}"` instead of `"{0}\r\n{1}\r\n{2}"` so that the error won't happen anymore.

TheRut:
Removed the try/catch. I did leave the update to the form layout so that the click-to-copy functionality captures the version information now (which I believe was the intent).

Barsonax:
Ok, will try to test this soon

ilexp (approved):
Works as expected. @Barsonax (Squash-)Merge as you see fit 👍

Barsonax:
Squash merged from mobile :)

Issue #784, opened 2020-03-23
Canvas.DrawImage


Shadowblitz16:
### Summary
Please add the following for easy image drawing through the canvas...
`void Canvas.DrawImage(Rect dst, Rect src, Material material, float rotation, ColorRgba blend);`
### Analysis

This allows for high level image drawing.
both rects would be in pixels

Pros
- Easy to draw images

ilexp:
Partially addressed in #784: There is now a high-level `canvas.DrawImage` method that provides a shortcut for quickly drawing a texture as-is, at a specified position. 

The more complex operations that you mentioned can already be achieved by using existing `Canvas` API and would lead to an explosion in the number of required overloads when incorporated directly into `DrawImage`, so they're not part of the API for now. This could be added later on, if needed, but would require additional design consideration.

Closing this for now, to be potentially re-evaluated as needed.

PR #878, opened 2020-10-17
Added Canvas.DrawImage (#784)

SirePi:
As per #784, this introduces a DrawImage(Material / BatchInfo) to the Canvas.
The code has been taken inspiration from FillRect, but feel free to add suggestions on how it can be improved


ChristianGreiner:
👍

Source/Core/Duality/Drawing/Canvas.cs
```diff
@@ -394,6 +394,168 @@ public void DrawThickLine(float x, float y, float x2, float y2, float width)
...
+			vertices[3].TexCoord = Vector2.UnitY;
+
+			vertices[0].Color = shapeColor;
+			vertices[1].Color = shapeColor;
+			vertices[2].Color = shapeColor;
+			vertices[3].Color = shapeColor;
+
+			this.State.SetMaterial(material);
+			this.State.TransformVertices(vertices, shapeHandle);
+			this.device.AddVertices(this.State.MaterialDirect, VertexMode.LineLoop, vertices, 4);
```
ilexp, line 436: Why is this using `LineLoop`? Wouldn't that mean only drawing a one-pixel outline instead of the image?

Source/Core/Duality/Drawing/Canvas.cs
```diff
@@ -394,6 +394,168 @@ public void DrawThickLine(float x, float y, float x2, float y2, float width)
...
+			vertices[3].TexCoord = Vector2.UnitY;
+
+			vertices[0].Color = shapeColor;
+			vertices[1].Color = shapeColor;
+			vertices[2].Color = shapeColor;
+			vertices[3].Color = shapeColor;
+
+			this.State.SetMaterial(batch);
+			this.State.TransformVertices(vertices, shapeHandle);
+			this.device.AddVertices(this.State.MaterialDirect, VertexMode.LineLoop, vertices, 4);
```
ilexp, line 517: See above

Source/Core/Duality/Drawing/Canvas.cs
```diff
@@ -394,6 +394,168 @@ public void DrawThickLine(float x, float y, float x2, float y2, float width)
...
+			this.PushState();
+
+			if (w < 0.0f) { x += w; w = -w; }
+			if (h < 0.0f) { y += h; h = -h; }
+
+			Vector3 pos = new Vector3(x, y, z);
+
+			Vector2 shapeHandle = pos.Xy;
+			float offset = this.State.DepthOffset;
+			ColorRgba shapeColor = material.Res.MainColor;
```
ilexp, line 411: The material color shouldn't go into the vertex color, since material parameters [are applied in the shader](https://github.com/AdamsLair/duality/blob/b31fbab9c85bf3e018dd9025f510f0ae25a19784/Source/Core/Duality/EmbeddedResources/Minimal.vert#L17). Instead, use the `ColorTint` property of the canvas state, [as seen in other Canvas methods](https://github.com/AdamsLair/duality/blob/b31fbab9c85bf3e018dd9025f510f0ae25a19784/Source/Core/Duality/Drawing/Canvas.cs#L941).

Source/Core/Duality/Drawing/Canvas.cs
```diff
@@ -394,6 +394,168 @@ public void DrawThickLine(float x, float y, float x2, float y2, float width)
 			this.DrawThickLine(x, y, 0, x2, y2, 0, width);
 		}
 
+		/// <summary>
+		/// Draws an image.
+		/// </summary>
+		public void DrawImage(ContentRef<Material> material, float x, float y, float z, float w, float h)
```
ilexp, line 400: This overload (and all others based on `Material` references) can just forward to the `BatchInfo` one and provide [material.Info](https://github.com/AdamsLair/duality/blob/b31fbab9c85bf3e018dd9025f510f0ae25a19784/Source/Core/Duality/Resources/Material.cs#L85) as a parameter.

Source/Core/Duality/Drawing/Canvas.cs
```diff
@@ -394,6 +394,168 @@ public void DrawThickLine(float x, float y, float x2, float y2, float width)
...
+		/// <param name="y"></param>
+		public void DrawImage(ContentRef<Material> material, float x, float y)
+		{
+			this.DrawImage(material, x, y, 0);
+		}
+
+		/// <summary>
+		/// Draws an image.
+		/// </summary>
+		public void DrawImage(BatchInfo batch, float x, float y, float z, float w, float h)
```
ilexp, line 481: To avoid an explosion in overload permutations with various parameters, I'd consider leaving out the simplified ones for the `BatchInfo` variant, since this is probably the rare one. We can always add them later, if needed.

ilexp, line 481: On a similar note, I think we should provide overloads taking a `ContentRef<Texture>` parameter, where we use [this.device.RentMaterial(...)](https://github.com/AdamsLair/duality/blob/b31fbab9c85bf3e018dd9025f510f0ae25a19784/Source/Core/Duality/Drawing/Canvas.cs#L334) to draw the texture with a hidden temp material. I imagine that this could be convenient for all cases where there isn't necessarily a full material defined somewhere.

ilexp (changes requested):

Source/Core/Duality/Drawing/Canvas.cs
SirePi: Aaand I apparently rolled back my commit.. first time I started from DrawRect by mistake and then something got messed up with my git repo. Thanks for the spot

Source/Core/Duality/Drawing/Canvas.cs
SirePi: will do

ilexp:
Thanks! Finally managed to look into this again and update the review.

Instead of requesting a few more changes, I took the time to add them myself right away, to avoid going stale again, since I think this is a pretty useful feature. Some of the changes were not part of earlier review requests and only occurred to me when doing a more thorough review today, checking out the most recent changes. I know it's not the cleanest way to go through with this PR, but right now it's the best I can manage, given the limited time I have - sorry about that, but I hope that's fine with you.

Changes I made and rationale behind them:

- Premise: `DrawImage` is a shortcut that aims specifically to fit the "I want to draw this texture as-is" use case, since all others are better off with existing, more complex / verbose API.
- Removed all `ContentRef<Material>` and `BatchInfo` overloads, since they are easily available via `canvas.State.SetMaterial` if needed.
- Removed all overloads that allow to specify with and height, since they're already available via `canvas.FillRect` when setting the material beforehand.
- Added a unit test for `DrawImage`.

ilexp (approved):

